PR #53, opened 2023-01-13
Fix workflow for release

Natoandro:


.github/workflows/release.yml
```diff
@@ -18,6 +18,9 @@ jobs:
       - uses: actions/checkout@v3
       - id: version
```
zifeo, line 19: You can even remove this step as it not used anymore

zifeo (approved):

PR #90, opened 2023-01-31
Met 75 check all todos fixme in the project

afmika:
This partially solves MET-75 (TODO / FIXME)


typegate/deno_import.json
```diff
@@ -20,6 +20,7 @@
     "oauth2_client": "https://deno.land/x/oauth2_client@v0.2.1/mod.ts",
     "test/mock_fetch": "https://deno.land/x/mock_fetch@0.3.0/mod.ts",
     "json_schema_typed": "https://deno.land/x/json_schema_typed@v8.0.0/draft_latest.ts",
-    "url_join": "https://deno.land/x/url_join@1.0.0/mod.ts"
+    "url_join": "https://deno.land/x/url_join@1.0.0/mod.ts",
+    "collections": "https://deno.land/std@0.139.0/collections/mod.ts"
```
Natoandro, line 24: We already have `"std/"` on line 3; or do we need features from that specific (old) version?

Natoandro: Imports could be:
```ts
import { mapValues } from "std/collections/map_values.ts";
```
This uses the version specified on the `"std/"` field, and download less files: only `map_values.ts` and its deps; `"collections/mod.ts"` would download all the imports in `mod.ts`.

typegate/src/engine.ts
```diff
@@ -31,6 +31,8 @@ import { Planner } from "./planner/mod.ts";
 import { FromVars } from "./runtimes/graphql.ts";
 import config from "./config.ts";
 import * as semver from "std/semver/mod.ts";
+import { mapValues } from "collections";
```
Natoandro, line 34: cf supra

typegate/src/runtimes/random.ts
```diff
@@ -60,33 +71,45 @@ export class RandomRuntime extends Runtime {
...
+              return this.chance.email();
+            }
+            return this.chance.string();
+          case "boolean":
+            return this.chance.bool();
+          case "array": {
+            const res = [];
+            let size = this.chance.integer({ min: 1, max: 10 });
+            while (size--) {
+              const childNodeName = this.getTgTypeNameByIndex(typ.items);
```
Natoandro, line 102: Calls `this.getTgTypeNameByIndex` on each iteration.

libs/xtask/src/codegen/typescript.rs
```diff
@@ -100,6 +100,11 @@ pub fn run() -> Result<()> {
 }
 
 fn print_module<W: Write>(cm: Lrc<SourceMap>, module: &Module, writer: W) -> Result<()> {
+    // ref: https://doc.rust-lang.org/std/env/consts/constant.OS.html
```
Natoandro, line 103: Though not critical, a compile type check would be more efficient.
```rust
#[cfg(target_os = "windows")]
let new_line = "\r\n";
#[cfg(not(target_os = "windows"))]
let new_line = "\n";
```

Natoandro (changes requested):

typegate/src/runtimes/random.ts
afmika: You are right, I will refactor fact.

Natoandro (dismissed):

Natoandro (dismissed):

Natoandro (approved):

PR #110, opened 2023-02-10
feat: add type node either

msga-mmm:
Add type node `either` which allows to create a JSON Schema with an `oneOf` field with the given variant types.


afmika (approved):
lgtm

Natoandro (approved):

PR #111, opened 2023-02-10
MET-67 graph implement full prisma mapping

afmika:
## This adds the following features into Prisma:
* aggregate (count, min, max, sum, avg)
* groupBy (count, min, max, sum, avg)
* pagination (skip, take)
* orderBy
* distinct (findMany)
* createMany
* deleteMany
* updateMany (set, multiply, divide, increment, decrement)
* upsert
* nested count (findMany, findUnique)
* Generalized where queries
* Generalized having queries for groupBy


typegate/tests/prisma/full_prisma_mapping_test.ts
```diff
@@ -0,0 +1,242 @@
...
+    `.expectData({
+      findManyPosts: [
+        { id: 10002, title: "Book 2" },
+        { id: 10003, title: "Book 3" },
+      ],
+    })
+      .on(e);
+  });
+
+  await t.should("orderBy likes and views", async () => {
```
Natoandro, line 100: I would include some entries with the same `likes` but different `views` to see if it is ordered correctly by `views`.

typegraph/typegraph/providers/prisma/type_generator.py
```diff
@@ -124,6 +122,42 @@ def get_where_type(
         else:
             return t.struct(fields).named(name)
 
+    def deep_map(self, tpe: t.typedef, fn: callable, seen=set()) -> t.struct:
```
Natoandro, line 125: What about `t.boolean`?

Natoandro (commented):
Nice :+1:

typegraph/typegraph/providers/prisma/type_generator.py
afmika: I have already removed the `if` condition, it was merely a test to see if it worked. Terminal nodes should be automatically handled in the last return statement (`t.boolean`, `t.string`, ..etc).

typegate/tests/prisma/full_prisma_mapping_test.ts
```diff
@@ -0,0 +1,242 @@
...
+  await t.should("do a count with findManyPost", async () => {
+    await gql`
+        query {
+          findManyUsers {
+            _count { posts extended_profile }
+          }
+        }
+    `.expectBody((body: any) => {
+      console.log("body findManyPost :::", body);
+    })
```
msga-mmm, line 66: Is this for debugging?

afmika: yes, it's still subject to change, the `_count` for `findUnique` is still awaiting for `t.union` to work properly or if I find another way of implementing it.

typegraph/typegraph/providers/prisma/runtimes/prisma.py
```diff
@@ -304,42 +308,150 @@ def executeRaw(self, query: str, *, effect: Effect) -> t.func:
...
         )
 
+    def aggregate(self, tpe: Union[t.struct, t.NodeProxy]) -> t.func:
+        self.__manage(tpe)
+        typegen = self.__typegen
+        _pref = get_name_generator("Aggregate", tpe)
+        tpe_nums = typegen.extract_number_types(tpe)
+        return t.func(
+            t.struct(
+                {
```
zifeo, line 364: I would suggest to add any arg type to the parameter of the function so user can override them if they want:

```
    def aggregate(self, tpe: Union[t.struct, t.NodeProxy], where=typegen.gen_query_where_expr(tpe)
                     .named(_pref("Where"))) -> t.func:
```

would allow 

```
db.aggregate(mytype, where=typegen.gen_query_where_expr(mytype, id=t.string().set(XXXX)))
```

typegate/src/typegraph/utils.ts
```diff
@@ -11,6 +11,19 @@ export function treeView(tg: TypeGraphDS, rootIdx = 0, depth = 4) {
     const edge = cyan(`${path.edges[path.edges.length - 1] ?? "[root]"}`);
     const idxStr = green(`${idx}`);
     console.log(`${indent}${edge} → ${idxStr} ${type.type}:${type.title}`);
+    if (type.type == "union") {
+      console.log(`${indent}anyOf → ids [${type.anyOf.join(", ")}]`);
```
zifeo, line 15: do not forget to use the logger with correct level if you want to keep the logs or remove the console log statement before the merge

zifeo (commented):
Nice!

PR #123, opened 2023-02-15
chore: update Cargo.lock

msga-mmm:
Outdated Cargo.lock is causing problems when installing meta-cli from source code.


zifeo:
lgtm

PR #139, opened 2023-02-20
chore: update Cargo.lock

msga-mmm:
Some prisma crate references are outdated, so an update to `Cargo.lock` is required.

This will also help to fix the problems in the Continuous Integration (CI) https://github.com/metatypedev/metatype/actions/runs/4219773704/jobs/7325454067


msga-mmm:
Pull request #140 resolves the dependency update better.

PR #148, opened 2023-02-23
MET-8: http runtime name clash

afmika:


zifeo (dismissed):

msga-mmm (approved):
It looks good to me.

Natoandro (changes requested):
See https://metatypedev.slack.com/archives/C03J804A7TQ/p1677264476707969

typegraph/typegraph/runtimes/http.py
```diff
@@ -85,6 +85,7 @@ class RESTMat(Materializer):
     content_type: str = field(kw_only=True, default="application/json")
     header_prefix: Optional[str] = "header#"
     query_fields: Optional[Tuple[str, ...]] = field(kw_only=True, default=None)
+    rename_fields: Optional[Tuple[Tuple]] = field(kw_only=True, default=None)
```
Natoandro, line 88: `Optional[Tuple[Tuple[str, str]]]`?

Natoandro (dismissed):

typegraph/typegraph/runtimes/http.py
Natoandro: I think a `dict` would be easier to write; and will not produce this kind of mistake: `(("field", "renamed"))`.
You can set `field(converter=frozendict)` to automatically transform it into `frozendict`.

afmika: You are right, I had errors with a raw dict as it is not hashable so I used tuple as a proof of concept. Thanks, I will try that.

PR #149, opened 2023-02-23
test(typegraph): add tests for arrays of optional values

msga-mmm:
These changes solve the Jira ticket MET-34.

The support for arrays of optional values was already implemented or working before this. I just have added a few tests to make sure it works properly plus a small fix to allow sending `null` values when passing an argument that is an array of optional values.

However, I'm not too sure if the implementation on the `nextBatcher` is good as I didn't quite understand the purpose or goal of this function.

```typescript
if (isArray(type)) {
  if (isOptional(this.type(type.items))) {
    return (x: any) => ensureArray(x);
  }
}
```


typegraph/typegraph/types.py
```diff
@@ -355,9 +355,22 @@ def _constraints(self):
...
+    of: TypeNode = field(default=None)
     default_value: Optional[str] = field(hash=False, default=None)
 
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+
+        of = self.of
+
+        if isinstance(of, NodeProxy):
+            of = of.get()
```
Natoandro, line 367: This would fail if the proxied type is not yet defined...

PR #150, opened 2023-02-23
feat: better logging and fix deployment

zifeo:
MET-115


afmika (approved):

typegate/tests/utils.ts
```diff
@@ -488,11 +491,13 @@ export async function recreateMigrations(engine: Engine) {
   const migrationsBaseDir = join(thisDir, "prisma-migrations");
 
   for await (const runtime of runtimes) {
+    console.log(runtime);
     const prisma = new PrismaMigrate(engine, runtime, null);
     const { migrations } = await prisma.create({
       name: "init",
       apply: true,
     } as any);
+    console.log("aaa");
```
Natoandro, line 500: Is this supposed to be here?

whiz.yaml
```diff
@@ -118,7 +118,7 @@ website_docgen:
         npx docusaurus docs:generate:graphql:prisma-migration
         npx docusaurus docs:generate:changelog
     depends_on:
-        - typegate1
```
Natoandro, line 121: Does not this depend on `typegate`?

Natoandro (approved):

Natoandro (dismissed):

whiz.yaml
zifeo: Typegate never finishes. I always thought that Whiz should have a way to manage those kind of task : e.g. if no stdout is detected after a while and the task is marked a "continuous".

Natoandro: You are right. Something we could implement is to check when there is a server listening on some port.

Natoandro (approved):

PR #195, opened 2023-03-10
Enable optional relationships

Natoandro:
MET-109


typegate/tests/prisma/self_relationship_test.ts
```diff
@@ -0,0 +1,143 @@
...
+        createNode: {
+          name: "Two",
+        },
+      })
+      .on(e);
+  });
+
+  await t.should("insert a list", async () => {
+    // This does not work!
+    // Should this work??
```
zifeo, line 61: I would say yes!

zifeo (dismissed):

typegate/tests/prisma/self_relationship_test.ts
```diff
@@ -0,0 +1,143 @@
...
+          next: {
+            name: "Second",
+          },
+        },
+      })
+      .on(e);
+  });
+
+  // TODO: enable; after implementation of "Full prisma mapping"
+  // t.should("find first nodes", async () => {
```
zifeo, line 124: Don't forget :)

typegate/tests/prisma/self_relationship_test.ts
Natoandro: I will see after we merge again the full prisma mapping...

typegraph/typegraph/providers/prisma/type_generator.py
```diff
@@ -327,9 +327,50 @@ def extract_number_types(self, tpe: t.struct) -> t.struct:
...
+            if isinstance(tpe, t.array):
+                nested_type = resolve_proxy(tpe.of)
+                if isinstance(nested_type, t.struct):
+                    return aggregate
+                else:
+                    # TODO: check prisma docs
+                    return sort
+
+            # scalar type
+            # ? are they all sortable??
```
afmika, line 362: I think the answer to this question would be yes (in most cases) since scalar types are either text or numbers (except arrays and BLOBs). But it can also depend on the target database. The Prisma engine should throw an error if it's not supported (in case the Typegraph allows it).

PR #147, opened 2023-02-22
MET-73: Make policies configurable per effect

Natoandro:

PR #199, opened 2023-03-10
Fix schema validation

afmika:
This fixes the schema generation and its validation.
Related to #111 and #147


typegate/src/planner/args.ts
```diff
@@ -816,14 +828,33 @@ class TypeMismatchError extends Error {
   }
 }
 
-const ajv = new Ajv();
+const visitedObject = new Set<string>();
```
zifeo, line 831: why is the cache global and not per instance of build?

typegate/src/planner/args.ts
```diff
@@ -832,29 +863,63 @@ class JsonSchemaBuilder {
   }
 
   build(typeNode: TypeNode): JSONSchema {
+    let resultSchema: JSONSchema = {};
```
zifeo, line 866: Is there a need to have this temporary variable instead of the return in all the cases?

zifeo (changes requested):
Great fix, some small style comments.

typegate/src/planner/args.ts
afmika: Because ajv is global, `visitedObject` is a subset of all schemas names present in the ajv instance above.

typegate/src/planner/args.ts
afmika: Yes, it helps handling `case Type.OPTIONAL` and `case Type.OBJECT` as I mutate the properties before a return. Possibly reducible to return statements but that would be more of an aesthetic issue I think as I'd have to create more local variables.

typegate/src/planner/args.ts
```diff
@@ -816,14 +828,33 @@ class TypeMismatchError extends Error {
...
+function refOf(typeNode: TypeNode): JSONSchema {
+  return { $ref: pathOf(typeNode) };
+}
+
+function addSchemaToAjv(schema: JSONSchema, path: string) {
+  if (ajv.getSchema(path)) return;
+  try {
+    ajv.addSchema(schema, path);
+    ajv.compile(schema); // cache if possible
+  } catch (_err) { /* ref cannot be resolved yet */ }
```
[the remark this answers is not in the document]
afmika: `Type.OBJECT` is the only type cached. In fact, it still works if I remove `ajv.compile(.)` in this function because all I have to do is to compile before the validation therefore if there are a references that cannot be resolved yet (eg. cross-refering types), the last compile statement right before `validator(value)` should throw.

typegate/src/planner/args.ts
```diff
@@ -109,7 +109,20 @@ export function collectArgs(
   const jsonSchema = buildJsonSchema(argTypeNode, typegraph);
   const validator = ajv.compile(jsonSchema);
   const validate = (value: unknown) => {
-    if (!validator(value) && validator.errors) {
+    let isValid = false;
+    try {
+      isValid = validator(value);
+    } catch (error) {
```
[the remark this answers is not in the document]
afmika: Not really, just a hint for the `FIXME`

typegate/src/planner/args.ts
```diff
@@ -109,7 +109,20 @@ export function collectArgs(
...
+      // FIXME: ajv "Maximum call stack size exceeded" error
+      // When the schema
+      // * is large, has multiple references
+      // * has nested either/union
+      // validator may stackoverflow even if the schema/value are
+      // not circular
+      // https://github.com/ajv-validator/ajv/issues/1581
```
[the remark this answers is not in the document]
afmika: I have tried `allErrors: true` and it seems to work (it logs warnings instead). I haven't tried `inlineRefs` yet, might worth a shot.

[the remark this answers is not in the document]
afmika: Same result, `inlineRefs` seems to only speed up the compilation.
https://ajv.js.org/options.html#inlinerefs

zifeo (approved):

typegate/src/planner/args.ts
zifeo: Hm, still not clear to me, what happens if a type called "Author" is in a given shape (let's call v1), and then in the same typegraph or another one, the same `wrapped.title` is referencing another shape (v2)? I understand that the type checking might end up wrong?

typegate/src/planner/args.ts
```diff
@@ -899,11 +996,17 @@ class JsonSchemaBuilder {
...
+      const currentNode = this.types[typeIdx];
+      return visitedObject.has(currentNode.title)
+        ? refOf(currentNode)
+        : this.build(currentNode);
+    });
+    // Ref:
+    // https://json-schema.org/understanding-json-schema/reference/type.html
     return {
-      // ...trimType(node),
+      type: this.listUnionEitherTypes(node, variantsKey),
```
Natoandro, line 1009: I think `type` is not necessary for `anyOf` and `oneOf`. Is there a specific reason for this?

typegraph/typegraph/providers/prisma/type_generator.py
```diff
@@ -124,6 +126,211 @@ def get_where_type(
...
+    def gen_query_where_expr(
+        self, tpe: t.struct, exclude_extra_fields=False
+    ) -> t.struct:
+        tpe = self.get_where_type(tpe)
+        extended_tpe = self.extend_terminal_nodes_props(tpe)
+        extended_tpe = rename_with_idx(extended_tpe, "extended_tpe")
+
+        # define the terminal expression
+        temp_props = {k: v.optional() for k, v in extended_tpe.props.items()}
+        intermediate = t.proxy("???")
```
Natoandro, line 183: ?

typegraph/typegraph/providers/prisma/utils.py
```diff
@@ -62,3 +72,14 @@ def only_unique(tpe: t.Type):
...
+
+
+# temporary redirect stdout to a file
+def stdout_to_file(obj: any, filename="stdout.txt"):
+    import sys
+
+    original_stdout = sys.stdout
+    with open(filename, "w") as f:
+        sys.stdout = f
+        print(obj)
```
Natoandro, line 84: `print(obj, file=f)`

typegraph/typegraph/types.py
```diff
@@ -504,10 +504,18 @@ class struct(typedef):
...
+    def detach(self):
+        return self.replace(
+            name=f"{self.type}_{self.graph.next_type_id()}", runtime=None
+        )
+
     def compose(self, props: Dict[str, typedef]):
         new_props = dict(self.props)
         new_props.update(props)
-        return self.replace(props=frozendict(new_props))
+        return self.replace(
```
Natoandro, line 516: I think `self.replace` should always replace the name...

Natoandro (commented):
Nice fix!

typegate/src/planner/args.ts
afmika: yes, when the root `anyOf`/`oneOf` is optional and contain nested `anyOf`/`oneOf`, the parent `anyOf`/`oneOf` must be aware of the types in the next `anyOf`/`oneOf` and so on. In simple terms, we cannot do `someAnyOfOneOf.type = ["null"]` or `someAnyOfOneOf.type = ["object", "null"]`, we have to enumerate all types before the last `null`.

typegraph/typegraph/providers/prisma/type_generator.py
afmika: `???` is a placeholder, can be anything else, I mutate it later to self-loop the expression definition thus allowing infinite nested expressions dynamically.

typegraph/typegraph/providers/prisma/utils.py
afmika: Agreed but I opted for this because it looks a lot more explicit.

typegraph/typegraph/types.py
afmika: I don't understand, if it should always replace the name when used then isn't this already ok ? Or will it do that automatically without having to explicitly give a new name ?

Natoandro (approved):

Natoandro (approved):

PR #210, opened 2023-03-15
Feat: Hooks

Natoandro:
MET-113, MET-122


typegate/src/hooks.ts
```diff
@@ -0,0 +1,67 @@
...
+export function registerHook(when: "onPush", handler: PushHandler): void;
+export function registerHook(when: "onInit", handler: InitHandler): void;
+export function registerHook(
+  when: keyof Hooks,
+  handler: Hooks[keyof Hooks][number],
+): void {
+  (hooks[when] as unknown[]).push(handler);
+}
+
+const previousVersions: Map<string, TypeGraph> = new Map();
```
zifeo, line 37: Should not this come from the global register instead? This would avoid having typegraph stored at multiple places.

typegate/src/engine.ts
```diff
@@ -40,54 +34,24 @@ const introspectionDefStatic = await Deno.readTextFile(
...
-    typegraph = migration.transform(typegraph);
-    currentVersion = migration.next;
-  }
-
-  return typegraph;
-}
-
 export const initTypegraph = async (
   payload: string,
+  sync: boolean, // redis synchronization?
```
zifeo, line 39: maybe choose a better name, e.g. "nopushhook" as it does not seem related to the sync?

typegate/src/runtimes/prisma.ts
```diff
@@ -37,6 +42,59 @@ function isPrismaOperationMat(mat: Materializer): mat is PrismaOperationMat {
   return mat.name === "prisma_operation";
 }
 
+export interface PrismaRuntimeDS extends Omit<TGRuntime, "data"> {
+  data: PrismaRuntimeData;
```
zifeo, line 46: Omit, then re-add the data field, why?

typegate/src/runtimes/prisma.ts
```diff
@@ -37,6 +42,59 @@ function isPrismaOperationMat(mat: Materializer): mat is PrismaOperationMat {
...
+}
+
+registerHook("onPush", async (typegraph) => {
```
zifeo, line 49: I believe you may need to introduce "push" parameter (from the CLI) as one may want to push without apply the migration (and handle migration separately)

typegate/src/runtimes/prisma.ts
```diff
@@ -37,6 +42,59 @@ function isPrismaOperationMat(mat: Materializer): mat is PrismaOperationMat {
...
+
+    const res = nativeResult(
+      await native.prisma_deploy({
+        datasource,
+        datamodel,
+        migrations: migrations!,
+      }),
+    );
+
+    // TODO: log back to the user
```
zifeo, line 89: Indeed, we should aggregate all results from hook (e.g. allowing to warn when there was an upgrade in the format).

libs/xtask/src/codegen/jsonschema.rs
```diff
@@ -1,12 +1,25 @@
...
 
 use anyhow::{Context, Result};
-use common::typegraph::Typegraph;
+use common::typegraph::{FunctionMatData, ModuleMatData, PrismaRuntimeData, Typegraph};
 use schemars::schema_for;
 use std::io::Write;
 use std::path::Path;
 use std::{env, fs};
 
+macro_rules! add_schema {
```
zifeo, line 10: 👍

meta-cli/src/typegraph/mod.rs
```diff
@@ -14,30 +17,24 @@ use walkdir::{DirEntry, WalkDir};
...
-use typescript::parser::{transform_module, transform_script};
+use postprocess::postprocess;
 
 pub type LoaderResult = HashMap<String, Result<Vec<Typegraph>>>;
 
 #[derive(Clone)]
 pub struct TypegraphLoader<'a> {
     skip_deno_modules: bool,
     ignore_unknown_file_types: bool,
+    deploy: bool, // target is deployment
```
zifeo, line 28: I would rather put an enum `target = Target.deployment or Target.Serialize` to explicite this parameter.

typegate/src/engine.ts
Natoandro: It indicates if the typegraph is being initialized from the redis synchronization or a direct push from the user...

typegate/src/graphql/graphql.ts
```diff
@@ -98,6 +103,7 @@ export function parseGraphQLTypeGraph(typegraph: TypeGraphDS) {
       ...rootNode,
       title: "Query",
       properties: queryProperties,
+      config: { ...(rootNode.config ?? {}), __namespace: true },
```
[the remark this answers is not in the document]
Natoandro: It was missing, though all the other namespaces have it. See "Mutation" just below.
It did not cause any problem since it is the root namespace, but I think we might need it in the future.

typegate/src/hooks.ts
Natoandro: Yes, that is right! This will not work at all since the next push might run on another instance.

The problem is that the hook might rewrite the typegraph:
- adding some computed data so it does get computed only once, before the redis synchronization;
- removing some extra data that was only needed for this hook;
- ...

Only the last version (throughout all the hooks) is stored in the global register and synchronized...

I was even thinking about storing multiple versions, one version per hook... but hard to synchronize...

typegate/src/engine.ts
Natoandro: "onPush" hooks will only run on one instance of the typegate: the one that received the typegraph directly from the user. The others (getting their versions from redis) will not run the hooks...

typegate/src/runtimes/prisma.ts
Natoandro: Their types were incompatible, TypeScript complained...

typegate/src/runtimes/prisma.ts
Natoandro: :+1:
I will add `--no-migrations` which will skip the addition of the migrations files in the serialized typegraph...

typegate/src/hooks.ts
zifeo: What if we do not compare to previous version? What workflow would not be possible using another strategy?

Natoandro: It is just for optimization: for example skip the migrations if the migrations files did not change. We can drop it...

PR #211, opened 2023-03-15
feat: MET-102 and small fixes

zifeo:


typegate/native/src/runtimes/prisma/engine.rs
```diff
@@ -16,18 +16,17 @@ pub async fn register_engine(datamodel: String, tg_name: String) -> Result<Strin
         ignore_env_var_errors: false,
     };
     let engine = super::engine_import::QueryEngine::new(conf)
-        .with_context(|| format!("Error while registering engine for typegraph {tg_name}"))?;
+        .with_context(|| format!("Error while registering engine {engine_name}"))?;
     engine.connect().await?;
-    let engine_id = format!("{tg_name}_{}", super::ENGINES.len() + 1);
```
zifeo, line 21: bug here, dash map.len is not atomic registering same engine multiple times on fast computer

afmika: Nice catch, that will *definitely* happen.

afmika (approved):

PR #224, opened 2023-03-20
feat: add python wasi runtime

zifeo:


typegate/src/runtimes/python_wasi/memory.ts
```diff
@@ -0,0 +1,102 @@
...
+          viewV.setUint8(j, bytes[j]);
+        }
+
+        view.setInt32(i * 2 * 4, ptrV, true);
+        view.setInt32(i * 2 * 4 + 4, bytes.length, true);
+      } else if (Number.isInteger(args[i])) {
+        const len = 4;
+        const ptrV = this.allocate(len);
+        const viewV = new DataView(this.memory.buffer, ptrV, len);
+        viewV.setInt32(0, args[i] as number, true);
```
afmika, line 46:
```suggestion
        viewV.setInt32(   i  /* not 0 ? */  , args[i] as number, true);
```

afmika, line 46: Suppose I have 2 consecutive integers in `args`.
```suggestion
        viewV.setInt32(0 /* why not i ?? */, args[i] as number, true);
```

typegate/tests/runtimes/python_wasi/python_wasi.py
```diff
@@ -0,0 +1,15 @@
...
+
+with TypeGraph("python_wasi") as g:
+    public = policies.public()
+    python = PythonWasiRuntime()
+
+    g.expose(
+        test=t.func(
+            t.struct({"a": t.string()}),
+            t.string(),
+            python.from_lambda(lambda x: x["a"]),
```
afmika, line 12: Nice (y)

typegate/src/runtimes/python_wasi/memory.ts
zifeo: each args I, is allocating its own buffer and for integer we need to store the value at byte position 0

PR #225, opened 2023-03-21
MET-101: fix google service discovery in the tutorial

afmika:


website/docs/tutorials/import-your-api-blocks/fcm.graphql
```diff
@@ -1,8 +1,24 @@
-mutation {
-  send_notification(parent: "firebase-project-id") {
-    message {
-      id
-      message
+mutation SendNotification {
+  send_notification(
```
zifeo, line 2: for example in the doc/example, let's always keep them the shortest/smallest possible so unfamiliar people do not get lost. Your change are great to test/make production call, but for the sake of the tutorial we will take shortcut whenever we can (e.g. no name of mutation, etc.)

zifeo (commented):
Can you rebase and a snapshot test for this google api?

website/docs/tutorials/import-your-api-blocks/fcm.graphql
afmika: Good point, got it.

PR #227, opened 2023-03-24
Fix various grammar and semantical issues.

freezy:
First time reading through the tutorial. I've made it until `adding-more-runtimes`.

As a general note, make it more clear when the reader is supposed to update something, versus FYI content. I've added some instructions where I thought that wasn't clear enough.

Secondly, it should be obvious which of the displayed source files are complete and can be entirely copy+pasted, as opposed to files where the reader needs to pick line by line (@zifeo is already aware of that).


freezy:
> Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/metatypedev/metatype?pullRequest=227) before we can accept your contribution.

LOL no :)

PR #262, opened 2023-04-06
Fix either schema

afmika:
Related to #243


typegate/src/runtimes/typegraph.ts
```diff
@@ -37,8 +37,20 @@ const SCALAR_TYPE_MAP = {
   "string": "String",
 };
 
+function generateCustomScalarName(type: TypeNode) {
+  if (isScalar(type)) {
+    const id = type.type;
+    // We get "Expected {name} to be a GraphQL Object type."
+    // without the "_" prefix
+    return "_" + id[0].toUpperCase() + id.slice(1);
```
zifeo, line 45:
```
`_${}${}`
```

typegate/src/runtimes/typegraph.ts
```diff
@@ -158,11 +170,29 @@ export class TypeGraphRuntime extends Runtime {
...
           (t) => t.type, // for scalars: one GraphQL type per `type` not `title`
-        ).map((type) => this.formatType(type, false, false));
+        );
+        const scalarTypes = distinctScalars.map((type) =>
+          this.formatType(type, false, false)
+        );
+        const customScalarTypes = distinctScalars.map((node) => {
+          const key = generateCustomScalarName(node);
+          const custom = {
+            title: key,
```
zifeo, line 187: all output fields should be lazy, e.g. `() => key`

typegate/src/runtimes/typegraph.ts
```diff
@@ -158,11 +170,29 @@ export class TypeGraphRuntime extends Runtime {
         };
 
         visitTypes(this.tg, getChildTypes(this.tg.types[0]), myVisitor);
-
-        const scalarTypes = distinctBy(
-          [...scalarTypeIndices].map((idx) => this.tg.types[idx]),
+        const distinctScalars = distinctBy(
+          [...scalarTypeIndices].map((idx) => {
+            const node = this.tg.types[idx];
+            this.scalarIndex.set(node.type, idx);
```
zifeo, line 176: this.customScalars looks to be only used locally, is there any need to have it class based?

typegate/src/runtimes/typegraph.ts
```diff
@@ -158,11 +170,29 @@ export class TypeGraphRuntime extends Runtime {
...
+        const customScalarTypes = distinctScalars.map((node) => {
+          const key = generateCustomScalarName(node);
+          const custom = {
+            title: key,
+            type: "object",
+            properties: {
+              [node.type]: this.scalarIndex.get(node.type),
+            },
+          } as ObjectNode;
+          this.customScalars.set(key, custom);
```
zifeo, line 193: This custom does not seem to at "the correct place", why not generating this payload later in the formatType function? This would avoid to "hack" a class attribute for that.

zifeo (changes requested):
Great job, small comments on the technical details.

typegate/src/runtimes/typegraph.ts
afmika: I was projecting on potentially re-use the types, it was also motivated by explicitly throwing a custom error in case it does not exist in the map.

typegate/src/runtimes/typegraph.ts
afmika: It's used in `formatType`

typegate/src/runtimes/typegraph.ts
afmika: I don't understand, `custom` is of type `ObjectNode`
https://github.com/metatypedev/metatype/blob/main/typegate/src/types/typegraph.ts#L86

typegate/src/runtimes/typegraph.ts
zifeo: `scalarIndex` seems to be local to this function or my browser search is buggy, `customScalars` is used in `formatType`

typegate/src/runtimes/typegraph.ts
zifeo: I would advise against keeping something in memory for a potential re-use (by another subsequent query).

When there is a class attribute, the user expect that this attributes are global to the "work scope" of the class, which is not the case: only when the materializer is "schema", this field is used (e.g. the field could be passed as an argument to the formatType function). Or better, the `custom` object creation can be done directly in formatType, avoiding having to keep in memory some we have prebuilt (and maybe that will not be used if the query do not ask for those fields).

typegate/src/runtimes/typegraph.ts
afmika: yes, that is the case.

afmika: `scalarIndex` will hold the number id of the scalar types as the name suggests, it is used in `formatType` and `getSchema`, then cleared after. I also removed the `customScalars` class attribute, items will be regenerated as we need them.

typegate/src/runtimes/typegraph.ts
```diff
@@ -339,46 +376,41 @@ export class TypeGraphRuntime extends Runtime {
...
+        const description = `${type.type} type\n${
+          Array.from(titles).join(", ")
+        }`;
 
-      for (const [field, idx] of objects) {
-        variantsAsObject.properties[field] = idx;
-      }
-      for (const [field, idx] of remaining) {
-        variantsAsObject.properties[field] = idx;
+        return {
```
zifeo, line 388: Can you add comment explain why we are making this trick for all cases?

afmika: When a either/union node is on the input, we make it so that it is seen as a custom scalar when introspected, the description will inform about the variants. 
When it is on the output, it is seen as a native `UNION`, but since `UNION` can only support objects, we have to map the scalar variants to a custom object, for example `int -> { title: "_Integer", fields: { integer: typeIdx }}`.

afmika: The trickiest part is handling scalars on the output, when we are sure that the processed type has union/either, we have to generate the custom object types beforehand (`hasUnion` flag), this way when an object is processed with `formatType`, we don't have to worry about the missing types anymore since we cannot refer to types that does not exist in the schema list.

typegate/src/runtimes/typegraph.ts
afmika: I still kept `scalarIndex` class property because it would require passing a scalarIndex to every `formatType` call, thus adding more complexity and more branching and more `undefined`s checks in case it is not used. `formatType` is just as useful as a helper function in a class in this context. The content is also cleared when all calls are terminated anyway.

zifeo (approved):

Natoandro (approved):

PR #267, opened 2023-04-12
MET-112: gate better error message for invalid operation type

afmika:


typegate/src/planner/mod.ts
```diff
@@ -140,18 +140,29 @@ export class Planner {
...
+        const proposition = closestWord(name, allProps);
         const formattedPath = this.formatPath(node.path);
-        throw new Error(
-          `'${name}' not found at '${formattedPath}', available names are: ${suggestions}`,
-        );
+        if (proposition) {
+          throw new Error(
+            `'${name}' not found at '${formattedPath}', did you mean '${proposition}' ?`,
+          );
+        } else {
```
zifeo, line 160: if () {
  throw ..
}
// no else continue directly

typegate/src/utils.ts
```diff
@@ -141,3 +142,23 @@ export function nativeVoid(res: "Ok" | { Err: { message: string } }): void {
...
+) {
+  const [top] = list
+    .map((word) => {
+      const s = ignoreCase ? str.toLowerCase() : str;
+      const t = ignoreCase ? word.toLowerCase() : word;
+      return { word, score: levenshtein(s, t) };
+    })
+    .sort((a, b) => a.score - b.score);
+
+  if (top.score > maxDistance) {
```
zifeo, line 160: top can be null if the list is empty

zifeo (commented):
Nice, + test 👍

typegate/src/utils.ts
afmika: Thanks, nice catch.

typegate/src/planner/mod.ts
```diff
@@ -470,4 +493,35 @@ export class Planner {
...
     return [this.operationName, ...path].join(".");
   }
+
+  /**
+   * Idea:
+   * query: [a1, a2, ..], mutation: [b1, a1, b2, ..]
+   * => a1: [query, mutation], a2: [query], ..., b1: [mutation] ...
+   */
+  private getReverseMapNameToQuery() {
+    const indices = ["query", "mutation"].map((label) =>
```
zifeo, line 503: I would generalize this to any types included in the `properties` of the root type. We may support some subscription in the future or even go beyond what graphql spec recommend.

typegate/src/planner/mod.ts
```diff
@@ -470,4 +493,35 @@ export class Planner {
...
+      this.tg.type(0, Type.OBJECT).properties?.[label]
+    );
+    const res = new Map<string, Set<string>>();
+    for (const idx of indices) {
+      const { fields, title } = this.collectFieldNames(idx);
```
zifeo, line 508: unless the function can be used elsewhere, I would inline it to avoid confusion with the usage of the function (and construction object to deconstruct it afterwards)

zifeo (changes requested):

typegate/src/planner/mod.ts
afmika: `getReverseMapNameToQuery` calls `collectFieldNames`, both only depend on `tg: Typegraph` so we can generalize, and move the definitions to `utils.ts` or inline above the class definition.

zifeo (approved):

PR #268, opened 2023-04-13
Improve CLI DX

Natoandro:
MET-105, MET-40, and follow-up for MET-53


whiz.yaml
```diff
@@ -4,12 +4,12 @@ env:
 meta:
   watch:
     - "meta-cli/src/**/*.rs"
-    - "examples/metatype.yaml"
     - "typegraph/typegraph/**/*.py"
   command: "cargo run --color always -p meta-cli -- -C examples dev"
   env:
     VIRTUAL_ENV: $PWD/examples/.venv
     PATH: $PWD/examples/.venv/bin:$PATH
+    CLICOLOR_FORCE: "1"
```
zifeo, line 12: why is not that always the default?

Natoandro: On whiz, stdout is piped so it is detected to be a non-terminal.

PR #271, opened 2023-04-14
MET-34: enable array of optional's

afmika:


typegate/tests/type_nodes/array_of_optional.py
```diff
@@ -0,0 +1,49 @@
...
+        }
+    )
+
+    rec = t.struct(
+        {
+            "struct_array": t.array(nested.optional()).optional(),
+            "string_array": t.array(t.string().optional()).optional(),
+            "integer_array": t.array(t.integer().optional()).optional(),
+            # TODO:
+            # should we support null value for enum/union/either arrays ?
```
msga-mmm, line 21: Not too sure if we should, but I think it's in the JSON Schema so maybe it'd be a good idea.

A naive implementation in the typegraph could be:

```python
class optional(typedef):
    def __attrs_post_init__(self):
        super().__attrs_post_init__()

        of = self.of

        if isinstance(of, NodeProxy):
            of = of.get()

        if of._enum is not None:
            # an optional type with an enum requires the enum to have the value
            # `null` as variant
            #
            # optional-enum(red, blue) -> optiona-enum(red, blue, null)
            object.__setattr__(self, "of", of.replace(enum=(*of._enum, None)))
```

typegate/tests/type_nodes/array_of_optional.py
```diff
@@ -0,0 +1,49 @@
...
+        {
+            "a": t.string(),
+            "b": t.integer(),
+            "c": t.struct({"c1": t.string(), "inner": t.array(t.string().optional())}),
+        }
+    )
+
+    rec = t.struct(
+        {
+            "struct_array": t.array(nested.optional()).optional(),
```
msga-mmm, line 17: I think we could also add a test with `t.optional(t.array())`, I think it's handled a little different from `t.array().optional` so maybe is good to also test it.

msga-mmm (approved):
It looks good to me.

zifeo (approved):

typegate/tests/type_nodes/array_of_optional.py
afmika: As far as I know, in both cases the optional node is visited first which means both generate the same schema.

msga-mmm: I see, you're right.

PR #276, opened 2023-04-18
MET-154: (gate) prisma runtime find does not support empty string

afmika:
Added a few tests to validate the Prisma Engine behavior, also filtered `id` and `unique` fields for `find` queries.


typegate/tests/prisma/prisma_edgecases.py
```diff
@@ -0,0 +1,30 @@
+from typegraph import TypeGraph, effects, policies, t
+from typegraph.providers.prisma.runtimes.prisma import PrismaRuntime
+
+with TypeGraph("prisma") as g:
+    # schema ref:
+    # https://www.prisma.io/docs/reference/api-reference/prisma-client-reference
+
```
zifeo, line 7: For later pull requests, try to come up with addition to current testing typegraphs instead of creating new ones as it might introduce a cost when we have to refactor something on all of them.

zifeo (approved):
Should it also add support for findFirst?

typegate/tests/prisma/prisma_edgecases.py
afmika: Sure, as for a reason, I created a new tg for simplification and to isolate the issue, all the typegraphs seemed too heavy and unrelated. I also expected `min(.).max(.)` schema issue to be a bug on our side thus potentially requiring further edgecase testings.

afmika: `findFirst` is not yet implemented but it should be easy to add support for it.

zifeo: Great, can you include it in this PR ?

afmika: no problem 👍

msga-mmm (dismissed):

zifeo (approved):

PR #290, opened 2023-04-28
Typecheck: custom validators

Natoandro:


typegate/native/src/typescript.rs
```diff
@@ -0,0 +1,31 @@
...
+    source: String,
+}
+
+#[deno]
+enum FormatCodeOut {
+    Ok { formatted_code: String },
+    Err { message: String },
+}
+
+// should this only added on debug build?
```
zifeo, line 19: No, we may need it with the console :)

typegate/src/validator/input/compiler.ts
```diff
@@ -0,0 +1,168 @@
...
+        const { collectErrorsIn: errVar = "errors", prepare = [] } = node;
+        let closeBrackets = 0;
+        let ifPrefix = "";
+        if (elseClause) {
+          if (prepare.length > 0) {
+            yield `else {`;
+            for (const prep of prepare) {
+              yield prep;
+            }
+            ++closeBrackets;
```
zifeo, line 64: Can we use += 1 everywhere so it is more readable?

zifeo:
@Natoandro closing this one?

PR #300, opened 2023-05-03
MET-148: Optimized typecheck

Natoandro:
This solves MET-148.
This is an alternative implementation to #290.
It uses recursive function calls since the inlining tried in #290 had a huge performance hit.
Anyway, the stack overflow has disappeared completely and performance is way much better.


zifeo (dismissed):

afmika (approved):
Very nice 👍

PR #304, opened 2023-05-12
Query result validator

Natoandro:
- Compiled validator
- Only include selected fields in the result (no additional step for removal)
- Selections for union of object types


typegate/src/engine.ts
```diff
@@ -285,14 +339,45 @@ export class Engine {
...
+          return variant;
+        });
+        for (const variant of distinct(resultVariants)) {
+          cache[`${stageId}$${variant}`] = [];
+          lenses[`${stageId}$${variant}`] = [];
+        }
+        for (const [i, variant] of resultVariants.entries()) {
+          cache[`${stageId}$${variant}`].push(cache[stageId][i]);
+          lenses[`${stageId}$${variant}`].push(lenses[stageId][i]);
+        }
```
afmika, line 377: I would suggest to refactor `(stageId, variant) -> string` to a function

typegate/src/typecheck/result.ts
```diff
@@ -0,0 +1,387 @@
...
+              cg.generateStringValidator(typeNode);
+              break;
+          }
+        }
+      } else {
+        // TODO: cannot check enum - perhaps we should disable enums for non-scalar types??
+
+        switch (typeNode.type) {
+          case "optional": {
+            const c = entry.selectionSet != null ? ++this.counter : null;
```
afmika, line 129: I think that's a good idea.

typegate/src/typecheck/common.ts
```diff
@@ -0,0 +1,61 @@
...
+      return true;
+    } catch (_e) {
+      return false;
+    }
+  },
+  email: validator.isEmail,
+  // TODO validatorjs does not have a URI validator, so this is stricter than expected
+  uri: (value: string) => {
+    console.log("validating uri", value);
+    return validator.isURL(value, {
```
afmika, line 35:
```typescript
return validator.isDataURI(value) || validator.isURL(value, ...)
```

afmika (commented):
Nice 👍

typegate/src/engine.ts
```diff
@@ -219,11 +235,14 @@ export class Engine {
     verbose: boolean,
   ): Promise<JSONValue> {
     const ret = {};
-    const cache: Record<string, unknown> = {};
-    const lenses: Record<string, unknown> = {};
+    const cache: Record<string, unknown[]> = {};
```
zifeo, line 238: why is an array required? maybe you can add comment on the example data cached and put in the lenses?

typegate/src/engine.ts
```diff
@@ -236,16 +255,51 @@ export class Engine {
...
       } = stage.props;
 
+      const level = path.length;
+      const stageId = stage.id();
+      // const parentId = parent.id();
+      const parentId = parent && stageId.slice(0, stageId.lastIndexOf("."));
+
+      let activeSelection = getActiveSelection();
+      while (activeSelection != null && level < activeSelection.level) {
+        selections.pop();
```
zifeo, line 268: Add a test for the while loop?

typegate/src/engine.ts
```diff
@@ -236,16 +255,51 @@ export class Engine {
...
+      }
+
+      if (activeSelection != null) {
+        if (parentId == null) {
+          throw new Error("unexpected state");
+        }
+
+        if (activeSelection.level === level) {
+          // potential variant transition
+          const match = parentId?.match(VARIANT_TRANSITION_REGEX);
```
zifeo, line 279: regex matching is costly, is there another alternative?

typegate/src/engine.ts
```diff
@@ -285,14 +339,45 @@ export class Engine {
...
       }
-      const field = path[path.length - 1] as any;
+
+      const field = path[path.length - 1];
       if (node !== "") {
         lens.forEach((l: any, i: number) => {
-          l[field] = res[i];
+          // Objects are replaced by empty objects `{}`.
+          // It will be populated by child compute stages using values in `cache`.
+          l[field] = withEmptyObjects(res[i]);
```
zifeo, line 348: if it is replaced anyway later, why putting empty object?

typegate/src/planner/mod.ts
```diff
@@ -109,90 +105,138 @@ export class Planner {
...
-        selection.map((n) => n.name?.value),
-        typ.type,
-        Object.entries(props).reduce(
-          (agg, [k, v]) => ({ ...agg, [k]: this.tg.type(v).type }),
-          {},
-        ),
-      );
+    if (selectionSet == null) {
+      if (this.isSelectionSetExpectedFor(typeIdx)) {
+        const path = this.formatPath(node.path);
```
zifeo, line 111: add a test for here?

typegate/src/planner/mod.ts
```diff
@@ -109,90 +105,138 @@ export class Planner {
...
+      const selections = selectionSet.selections.map((sel) => {
+        if (sel.kind !== Kind.INLINE_FRAGMENT || sel.typeCondition == null) {
+          throw new Error("Expected inline fragment with type condition");
         }
-        // if the above fails, tell the user in case they made a typo
-        const suggestion = closestWord(name, allProps);
-        if (suggestion) {
+        const typeName = sel.typeCondition.name.value;
+        if (!unselectedVariants.has(typeName)) {
+          const path = this.formatPath(node.path);
```
zifeo, line 160: test for here?

typegate/src/planner/mod.ts
```diff
@@ -501,4 +504,73 @@ export class Planner {
   private formatPath(path: string[]) {
     return [this.operationName, ...path].join(".");
   }
+
+  private isSelectionSetExpectedFor(typeIdx: number): boolean {
+    const typ = this.tg.type(typeIdx);
+    if (typ.type === Type.OBJECT) {
+      return true;
```
zifeo, line 511: test for the next three code cov warning?

typegate/src/typecheck/matching_variant.ts
```diff
@@ -0,0 +1,137 @@
...
+        continue;
+      }
+      const typeNode = this.tg.type(typeIdx);
+
+      if (typeNode.enum != null) {
+        cg.generateEnumValidator(typeNode);
+      } else {
+        switch (typeNode.type) {
+          case "boolean":
+            cg.generateBooleanValidator(typeNode);
```
zifeo, line 80: test?

typegate/src/typecheck/matching_variant.ts
```diff
@@ -0,0 +1,137 @@
...
+          case "object":
+            cg.generateObjectValidator(
+              typeNode,
+              mapValues(typeNode.properties, functionName),
+            );
+            queue.push(...Object.values(typeNode.properties));
+            break;
+          case "union":
+            cg.generateUnionValidator(
+              typeNode,
```
zifeo, line 106: test for the next warnings?

typegate/src/typecheck/result.ts
zifeo: can we avoid ++ and use += 1 for clarity?

typegate/src/typecheck/result.ts
```diff
@@ -0,0 +1,387 @@
...
+      }
+
+      case Kind.INLINE_FRAGMENT: {
+        if (node.typeCondition != null) {
+          throw new Error("Unexpected type condition on non-union type");
+        }
+        return node.selectionSet.selections.flatMap((selectionNode) =>
+          this.getChildEntriesFromSelectionNode(
+            typeNode,
+            entry,
```
zifeo, line 290: test?

zifeo (commented):
very nice!

typegate/src/typecheck/result.ts
Natoandro: > I don't quite follow, aren't `t.enum` variants `str` only ?

`t.enum()` is an alias to `t.string().enum()`. Any type is valid.

typegate/src/engine.ts
Natoandro: Objects always start empty. The fields are set by child compute stages (corresponding to the field selections in the query).

zifeo: Yes, but letting `l[field] = res[I]` would avoid creating new object and still do the same behaviour? In this part of the engine, we should avoid creating or manipulating object/values so we have the less overhead possible.

Natoandro: The object might have fields that are not  selected by the query. Replacing with empty object will ensure that we only have the selected fields in the final result.

zifeo (approved):

PR #311, opened 2023-05-16
fix: all importers

afmika:


zifeo (approved):
image
were deleted but not replaced?

Natoandro (dismissed):

afmika:
> image were deleted but not replaced?

These files were generated in the wrong place, I kept the generated snapshots that align with the ci.

zifeo (approved):

PR #312, opened 2023-05-17
feat: File upload

Natoandro:
Solve MET-173.


libs/common/src/typegraph/types.rs
```diff
@@ -125,6 +125,15 @@ pub struct StringTypeData {
     pub format: Option<StringFormat>,
 }
 
+#[cfg_attr(feature = "codegen", derive(JsonSchema))]
+#[skip_serializing_none]
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct FileTypeData {
+    pub min_size: Option<u32>,
+    pub max_size: Option<u32>,
```
zifeo, line 134: add allowed mime-types?

typegate/native/src/runtimes/s3.rs
```diff
@@ -16,6 +19,22 @@ struct S3Client {
     endpoint: String,
 }
 
+fn create_bucket(client: S3Client, bucket: &str) -> Result<Bucket, S3Error> {
```
zifeo, line 22: rename to get_bucket? the current name may indicate this indeed creates a bucket

typegate/native/src/runtimes/s3.rs
```diff
@@ -156,3 +151,71 @@ fn s3_list(client: S3Client, bucket: &str, path: &str) -> S3Items {
...
+        Ok(bucket) => bucket,
+        Err(e) => {
+            error!("{e:?}");
+            return S3PresigningOut::Err {
+                message: e.to_string(),
+            };
+        }
+    };
+
+    bucket.set_path_style();
```
zifeo, line 210: this should be optional to support both native s3 and s3 compatible backend

typegate/native/src/runtimes/s3.rs
```diff
@@ -156,3 +151,71 @@ fn s3_list(client: S3Client, bucket: &str, path: &str) -> S3Items {
...
+
+    match bucket.presign_get(presigning.key, presigning.expires, None) {
```
zifeo, line 212: maybe expose the other parameters that the pre-signing require?

typegate/tests/graphql/request_parser_test.ts
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  await t.should("successfully query with file", async () => {
+    const fileContent = "Hello, World!";
+    const q = gql`
+        mutation($file: File!) {
+          singleUpload(file: $file) { url }
+        }
+      `
+      .withVars({
```
zifeo, line 24: make withVars check the file type instead of have a withFile? we will likely publish the testing utilities as a separate package in a few weeks so we should make the easiest interface possible

typegate/tests/utils/metatest.ts
```diff
@@ -158,7 +158,13 @@ export class MetaTest {
   ): Promise<boolean> {
     const res = await this.t.step({
       name: `should ${fact}`,
-      fn,
+      fn: async (t) => {
+        try {
+          await fn(t);
+        } catch (e) {
+          console.error(e);
```
zifeo, line 165: is this enough to fail the test or rethrow is needed?

typegate/tests/runtimes/s3/s3_test.ts
```diff
@@ -0,0 +1,64 @@
...
+      TG_S3_TEST_ACCESS_KEY: "minio",
+      TG_S3_TEST_SECRET_KEY: "password",
+    },
+  });
+
+  const fileContent = "Hello, World!";
+
+  await t.should("upload file successfully", async () => {
+    await gql`
+      mutation ($file: File!, $path: String!) {
```
zifeo, line 21: Should not we add a custom scalar called File for the introspection?

typegate/src/runtimes/s3.ts
```diff
@@ -111,6 +113,34 @@ export class S3Runtime extends Runtime {
...
+              bucket as string,
+              path as string,
+              new Uint8Array(await (file as File).arrayBuffer()),
+            ),
+          );
+          return key;
+        };
+      }
+
+      if (name === "download_url") {
```
zifeo, line 131: We likely need to support two mats:
1. presigned with some more parameters (see official s3 doc)
2. canonical or classical url which might be used with or without the bucket name (e.g. only the s3 key/path)

typegate/src/runtimes/s3.ts
```diff
@@ -111,6 +113,34 @@ export class S3Runtime extends Runtime {
...
 
+      if (name === "upload") {
+        const { bucket } = matData;
+        return async ({ file, path }) => {
+          const { key } = nativeResult(
+            await native.s3_upload(
+              this.client,
+              bucket as string,
+              path as string,
+              new Uint8Array(await (file as File).arrayBuffer()),
```
zifeo, line 124: how about multiple (named) files?

zifeo (dismissed):

typegate/native/src/runtimes/s3.rs
```diff
@@ -156,3 +151,71 @@ fn s3_list(client: S3Client, bucket: &str, path: &str) -> S3Items {
...
+        Err(e) => {
+            return S3UploadedFile::Err {
+                message: e.to_string(),
+            }
+        }
+    };
+
+    bucket.set_path_style();
+
+    let temp = RT.block_on(bucket.put_object(key, file));
```
zifeo, line 179: Side note: I used the s3 runtime in Rust because the deno one was not complete. Now that we can use the one of node, maybe the easier would be keep it in Deno. This is would avoid to load the entire file into memory (when passing from Deno to Rust, and make use of stream to upload the file).

afmika (approved):
👍

PR #315, opened 2023-05-21
feat: ignore files without tg (MET-175) + temporal fixes

zifeo:


meta-cli/src/typegraph/loader/discovery.rs
```diff
@@ -111,63 +96,61 @@ impl FileFilter {
...
             },
         );
+        let ignore = config.base_dir.join(".gitignore");
+        let gitignore = if ignore.exists() {
+            Some(Gitignore::new(ignore).0)
+        } else {
+            None
+        };
+        let matcher = RegexMatcher::new_line_matcher("with TypeGraph")?;
+
```
afmika, line 106: I would suggest `with(\s+)TypeGraph`. We never know.

afmika (approved):
Nice

Natoandro (approved):

PR #318, opened 2023-05-23
fix: add explicit error message

afmika:
Solves MET-137


zifeo (dismissed):

typegate/tests/utils/q.ts
```diff
@@ -264,26 +277,54 @@ export class Q {
       defaults["Authorization"] = await this.contextEncoder(context);
     }
 
+    const clean = (headers: Record<string, string>) => {
+      for (const [k, v] of Object.entries(headers)) {
+        if (v === "NULL") {
```
zifeo, line 282: when (or why) does that happen?

zifeo (approved):

typegate/tests/utils/q.ts
afmika: NULL is a placeholder value, it can be anything else. It is set when `withoutHeaders(headers: Array<string>)` is used.

PR #319, opened 2023-05-23
fix: Python typing error happening in `example.py` file

Kazy:
This fixes a few typing issues required to make the sure the `example.py` file we get when following the tutorial doesn't raise any error with Pyright.


zifeo (approved):
Nice, thanks!

Natoandro (approved):

Kazy:
Removed the unused import and fixed the type of `default_policy` which could actually also be a list of policies.

PR #321, opened 2023-05-24
feat(python): update types to make sure the tutorial works with pyright

Kazy:
Some more coming from the GraphQL part. We can wait for me to complete the tutorial before merging this one, in case I need to do some additional modifications.


Natoandro (approved):

zifeo:
thanks! Closing because typegraph SDK has been redesigned and we integrate better typing now :)

PR #327, opened 2023-05-29
feat: remove cookie auth, fix typegraph detection, fix injection, add auth doc, add regression test for doc

zifeo:


afmika (approved):
Nice 👍

PR #337, opened 2023-06-12
feat: New typegraph SDK

Natoandro:
Solves [MET-178](https://metatype.atlassian.net/browse/MET-178).

- Typegraph definition in TypeScript/Deno and in Python


.github/workflows/tests.yml
```diff
@@ -140,6 +140,19 @@ jobs:
...
+      - changes
+    if: ${{ needs.changes.outputs.typegraph == 'true' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: ${{ runner.os }}-rust-${{ hashFiles('**/rust-toolchain.toml') }}-${{ hashFiles('**/Cargo.lock') }}
+      - run: cargo test --package typegraph_core -- --test-threads=1
```
zifeo, line 154: multi-thread breaks?

libs/common/src/typegraph/types.rs
```diff
@@ -92,11 +92,11 @@ pub struct NumberTypeData {
 #[derive(Serialize, Deserialize, Clone, Debug)]
 #[serde(rename_all = "camelCase")]
 pub struct IntegerTypeData {
-    pub minimum: Option<i64>,
-    pub maximum: Option<i64>,
-    pub exclusive_minimum: Option<i64>,
-    pub exclusive_maximum: Option<i64>,
-    pub multiple_of: Option<i64>,
+    pub minimum: Option<i32>,
```
zifeo, line 95: I recall i64 is limited by wasm, add maybe a comment to explain the i32 choice here?

meta-cli/src/typegraph/loader/mod.rs
```diff
@@ -123,7 +205,38 @@ impl Loader {
...
+                let mut command = Command::new("python3");
+                command
+                    .arg(path.to_str().unwrap())
+                    .envs(vars)
+                    .env("PYTHONUNBUFFERED", "1")
+                    .env("PYTHONDONTWRITEBYTECODE", "1");
+                command
+            }
+            ModuleType::Deno => {
+                let mut command = Command::new("deno");
```
zifeo, line 231: The typescript SDK will be also published on NPM, so we may need to have a fallback on node as loader (should be detectable based on the import format)

typegraph/core/src/conversion/runtimes.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+}
+
+fn convert_effect(eff: WitEffect) -> Effect {
+    match eff {
+        WitEffect::None => effect(EffectType::None, true),
+        WitEffect::Create(idemp) => effect(EffectType::Create, idemp),
+        WitEffect::Update(idemp) => effect(EffectType::Update, idemp),
+        WitEffect::Delete(idemp) => effect(EffectType::Delete, idemp),
+        WitEffect::Upsert(idemp) => effect(EffectType::Upsert, idemp),
```
zifeo, line 25: you can remove upsert (see MET-196)

typegraph/core/src/conversion/runtimes.rs
```diff
@@ -0,0 +1,105 @@
...
+use serde_json::json;
+
+fn effect(typ: EffectType, idempotent: bool) -> Effect {
+    Effect {
+        effect: Some(typ),
+        idempotent,
+    }
+}
+
+fn convert_effect(eff: WitEffect) -> Effect {
```
zifeo, line 19: implementing into may be easier for all types?

typegraph/core/src/conversion/runtimes.rs
```diff
@@ -0,0 +1,105 @@
...
+        WitEffect::Upsert(idemp) => effect(EffectType::Upsert, idemp),
+    }
+}
+
+pub fn convert_materializer(
+    c: &mut TypegraphContext,
+    s: &Store,
+    mat: &RawMaterializer,
+) -> Result<Materializer> {
+    match &mat.data {
```
zifeo, line 34: I wonder the [enum_dispatch](https://github.com/metatypedev/metatype/blob/main/meta-cli/src/cli/mod.rs#L82) pattern may not be easier than a single convert function

typegraph/core/src/global_store.rs
```diff
@@ -0,0 +1,149 @@
...
+};
+
+#[derive(Default)]
+pub struct Store {
+    pub types: Vec<T>,
+    pub type_by_names: HashMap<String, TypeId>,
+
+    pub runtimes: Vec<Runtime>,
+    pub materializers: Vec<Materializer>,
+    default_deno_runtime: Option<RuntimeId>,
```
zifeo, line 26: That would be great to add the typegraph spec (see use case here: https://github.com/metatypedev/metatype/pull/338/files#diff-a81d5a26c45ae8197d6468544ed183a9805b40c7d9e53dd1972ef322e8e0101aR233-R241)

typegraph/core/src/global_store.rs
```diff
@@ -0,0 +1,149 @@
...
+            Some(runtime) => runtime,
+            None => {
+                let runtime = self.register_runtime(Runtime::Deno);
+                self.default_deno_runtime = Some(runtime);
+                runtime
+            }
+        }
+    }
+
+    pub fn get_predefined_deno_function(&mut self, name: String) -> Result<MaterializerId> {
```
zifeo, line 114: can you explain the end usage of this one?

typegraph/core/src/runtimes.rs
```diff
@@ -0,0 +1,102 @@
...
+};
+
+type Result<T, E = TgError> = std::result::Result<T, E>;
+
+#[derive(Debug)]
+pub enum Runtime {
+    Deno,
+}
+
+#[derive(Debug)]
```
zifeo, line 17: should not this file be merged with common at one point? or some of the definitions?

typegraph/core/src/typegraph.rs
```diff
@@ -1,31 +1,218 @@
...
+        types: vec![Some(TypeNode::Object {
+            base: gen_base(params.name),
+            data: ObjectTypeData {
+                properties: IndexMap::new(),
+                required: vec![],
+            },
+        })],
+        runtimes: Vec::new(),
+        materializers: Vec::new(),
+        mapping: Default::default(),
```
zifeo, line 80: ...Default::default()?

typegraph/core/src/types.rs
```diff
@@ -2,42 +2,111 @@
...
-pub trait TypeFun {
-    fn getattr(&self, field: String) -> Option<Tpe>;
+#[allow(clippy::derivable_impls)]
+impl Default for TypeBase {
+    fn default() -> Self {
+        Self { name: None }
+    }
 }
 
+#[derive(Debug)]
```
zifeo, line 15: are tuples more convient than struct with fields?

typegraph/core/wit/typegraph.wit
```diff
@@ -1,28 +1,101 @@
...
+        delete(idempotency),
+        upsert(idempotency),
+    }
+
+    record base-materializer {
+        runtime: runtime-id,
+        effect: effect,
+    }
+
+    record materializer-deno-func {
```
zifeo, line 76: I believe we shall add high-level comments on this file for all types of objects that a developer may want to integrate in his own SDK

typegraph/tests/test_sdks.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def serialize(path: Path) -> str:
+    print(f"serializing {path}...")
+    proc = subprocess.run(
+        [
+            "cargo",
+            "run",
+            "-p",
+            "meta-cli",
```
zifeo, line 16: would be great if the SDK tests are controlled/written in Rust, so no multiple test framework to maintain or run (but if this adds too much overhead because of the types), we may fallback on all deno tests instead.

typegraph/tests/typegraphs/types.ts
```diff
@@ -0,0 +1,34 @@
...
+}, { name: "User" });
+
+const post = t.struct({
+  id: t.integer(),
+  author: user,
+}, { name: "Post" });
+
+typegraph("test-types", (g) => {
+  g.expose({
+    one: DenoRuntime.func(s1, b, {
```
zifeo, line 22: I believe even though there is only one DenoRuntime, we shall stick with the same declaration process as others:

```
deno = DenoRuntime() # always return singleton
deno.func()
```

zifeo (commented):
Super!

.github/workflows/tests.yml
Natoandro: Yes. All the types are stored in a global store, so the indexes are different each time if other threads register some types at the same time.

Natoandro: I think we could use thread_local for the global store instead of Mutex. I think it is not likely for a user to generate a typegraph from different threads. WDYT?

meta-cli/src/typegraph/loader/mod.rs
Natoandro: That would be JavaScript then instead of TypeScript? Or shall we transpile the code on our own?

.github/workflows/tests.yml
zifeo: Good idea!

typegraph/core/src/runtimes.rs
Natoandro: Yes. I think most of the types should be shared. But for now, I do not want to change a lot of things in `common`. I would be a lot easier to merge the we add more runtime supports in the new typegraph sdk.

typegraph/core/src/global_store.rs
Natoandro: Create a materializer for a predefined function. It is to prevent having more than one materializers for the same function.

meta-cli/src/typegraph/loader/mod.rs
Natoandro: I meant, the typegraph definition module.

zifeo (approved):

typegraph/core/Cargo.toml
```diff
@@ -13,7 +13,18 @@ enum_dispatch = "0.3.11"
 once_cell = "1.18.0"
 serde = { version = "1.0.164", features = ["derive"] }
 serde_json = "1.0.97"
-wit-bindgen = "0.7.0"
+# version has been bumped since python package wasmtime v9.0.0 does not
+# support component version 3
```
zifeo, line 17: can you share the tracking issue on Slack ?

zifeo (approved):

afmika (approved):
Very nice 👍

PR #344, opened 2023-06-22
fix: Fix migrations for target with prefix

Natoandro:
Solve [MET-198](https://metatype.atlassian.net/browse/MET-198)
* Use the same migration files for the same typegraph either the target uses a prefix or not.


meta-cli/src/config.rs
```diff
@@ -247,8 +221,11 @@ impl Config {
         }
     }
 
-    pub fn node(&self, profile: &str) -> &NodeConfig {
-        self.typegates.get(profile).unwrap_or(&DEFAULT_NODE_CONFIG)
+    pub fn node(&self, args: &CommonArgs, default_target: &str) -> NodeConfig {
```
zifeo, line 224: what if default target does not exist?

typegate/src/engine.ts
```diff
@@ -147,11 +151,12 @@ export class Engine {
...
     introspectionDefPayload: string | null = introspectionDefStatic,
   ) {
-    const typegraph = JSON.parse(payload);
-    const typegraphName = typegraph.types[0].title;
+    const typegraph: TypeGraphDS = JSON.parse(payload);
+    const typegraphName = TypeGraph.formatName(typegraph);
     response.typegraphName(typegraphName);
 
-    const secretManager = new SecretManager(typegraphName, secrets);
+    // not prefixed!
```
zifeo, line 158: 👍

meta-cli/src/cli/deploy.rs
```diff
@@ -154,7 +150,7 @@ impl Deploy<DefaultModeData> {
             );
         }
 
-        let node_config = config.node(&options.target).with_args(&deploy.node);
+        let node_config = config.node(&deploy.node, "deploy");
```
zifeo, line 153: I would only put a default target as dev for dev, none for deploy

meta-cli/src/cli/prisma.rs
```diff
@@ -179,9 +174,7 @@ impl Action for Deploy {
             .prisma
             .fill(&config)?
             .ok_or_else(|| anyhow!("No migrations in the migration directory"))?;
-
-        let node_config = config.node("deploy");
-
+        let node_config = config.node(&self.node, "deploy");
```
zifeo, line 177: same yet, not convinced of the deploy as default, rather nothing

zifeo (changes requested):

zifeo (approved):

PR #345, opened 2023-06-23
feat(Deno): support deno function defined in multiple files

afmika:
This implements MET-135


typegate/src/runtimes/deno/deno_messenger.ts
```diff
@@ -31,8 +31,8 @@ export class DenoMessenger extends LazyAsyncMessenger<Worker, Task, unknown> {
               ...permissions,
               // non-overridable permissions (security between typegraphs)
               run: false,
-              read: false,
-              write: false,
+              read: "inherit",
```
zifeo, line 34: give explicit read access to the tmp folder, otherwise typegraph can read other typegraph

typegate/src/runtimes/deno/deno_messenger.ts
```diff
@@ -31,8 +31,8 @@ export class DenoMessenger extends LazyAsyncMessenger<Worker, Task, unknown> {
...
+              write: "inherit",
```
zifeo, line 35: should be false, we disallow writing as it could overload the disk

typegate/tests/runtimes/deno/deno_test.ts
```diff
@@ -108,54 +107,28 @@ test("Deno runtime: permissions", async (t) => {
   });
 });
 
-test("Deno runtime: reloading", async (t) => {
-  const tmpDir = join(testDir, "tmp");
```
zifeo, line 112: shouldn't we keep this test?

afmika: The dynamic reload test creates a tmp file and its content is rewritten at runtime: this is impossible to 100% replicate in the current implementation as `scripts/deno/*` is sent to typegate then DenoRuntime imports from that.
Alternartively, we could rewrite any script at `./tmp/{hash}/*` and reproduce the reload (=script overwritten at runtime as of my understanding) but I am not quite sure what behavior it tests exactly.

meta-cli/src/cli/serialize.rs
```diff
@@ -65,12 +65,7 @@ impl Action for Serialize {
             loader = loader.with_postprocessor(postprocess::EmbedPrismaMigrations::default());
         }
 
-        #[cfg(feature = "typegraph-next")]
```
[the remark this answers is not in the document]
afmika: Deno postprocess is also used for generating the tarball in this implementation so I removed it.

meta-cli/src/typegraph/postprocess.rs
```diff
@@ -152,9 +195,9 @@ pub mod deno_rt {
                     continue;
                 };
 
-                let path = tg.path.as_ref().unwrap().parent().unwrap().join(path);
-                let code = std::fs::read_to_string(&path)?;
-                mat_data.code = transform_module(code)?;
+                let path = Path::new(path).to_owned();
```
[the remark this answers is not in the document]
afmika: `path` is the absolute path to the script entry point, the prefix `file:` is already removed at this line.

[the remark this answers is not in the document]
zifeo: I meant [this](https://github.com/metatypedev/metatype/pull/344) prefix

typegate/tests/runtimes/deno/deno_test.ts
zifeo: We shall manage to reproduce the same logic, by re-uploading a typegraph with a slight different function logic. That will allow us to test well the clean-up and re-setting process of the files.

libs/common/src/archive.rs
```diff
@@ -35,3 +36,40 @@ pub fn archive<P: AsRef<Path>>(folder: P) -> Result<Option<String>> {
...
+                        .context("Adding file to tarball")?;
+                }
+                count += 1;
+                Ok(())
+            }
+            Err(e) => Err(e),
+        }?;
+    }
+
+    if count > 0 {
```
zifeo, line 69: why is the count required? this is not supported by the tar?

meta-cli/src/typegraph/postprocess.rs
```diff
@@ -111,7 +149,9 @@ pub mod deno_rt {
                 let mut mat_data: ModuleMatData = object_from_map(std::mem::take(&mut mat.data))?;
                 if !mat_data.code.starts_with("file:") {
                     // TODO check imported functions exist
-                    mat_data.code = transform_module(mat_data.code)?;
+                    let path = mat_data.code.strip_prefix("file:").unwrap();
```
zifeo, line 152: @Natoandro is there a way to remove the file?

typegate/deno.json
```diff
@@ -34,6 +34,8 @@
     "levenshtein": "https://deno.land/x/levenshtein@v1.0.1/mod.ts",
     "aws-sdk/client-s3": "https://esm.sh/@aws-sdk/client-s3@3.335.0?pin=v124",
     "aws-sdk/s3-request-presigner": "https://esm.sh/@aws-sdk/s3-request-presigner@3.335.0?pin=v124",
-    "outdent": "https://deno.land/x/outdent@v0.8.0/mod.ts"
+    "outdent": "https://deno.land/x/outdent@v0.8.0/mod.ts",
+    "tar": "https://deno.land/std@0.128.0/archive/tar.ts",
+    "conversion": "https://deno.land/std@0.153.0/streams/conversion.ts"
```
zifeo, line 39: old old std lib → we are at 0.192.0

typegate/src/runtimes/deno/deno_messenger.ts
```diff
@@ -31,7 +31,7 @@ export class DenoMessenger extends LazyAsyncMessenger<Worker, Task, unknown> {
               ...permissions,
               // non-overridable permissions (security between typegraphs)
               run: false,
-              read: false,
+              read: [".", "tmp/scripts/deno"], // cwd for import paths
```
zifeo, line 34: why is "." required? can you customise the path per typegraph/runtime otherwise the different runtimes will be able to run others which should not be accepted

typegate/src/utils.ts
```diff
@@ -214,3 +225,82 @@ export function collectFieldNames(tg: TypeGraph, typeIdx: number) {
...
+  const buffer = base64.decode(tarb64);
+  const streamReader = new Blob([buffer])
+    .stream()
+    .pipeThrough(new DecompressionStream("gzip"))
+    .getReader();
+  const denoReader = conversion.readerFromStreamReader(streamReader);
+
+  const untar = new Untar(denoReader);
+  const entries = [];
+  for await (const entry of untar) {
```
zifeo, line 243: Does not untar.extract() do the same operation?

typegate/src/utils.ts
```diff
@@ -214,3 +225,82 @@ export function collectFieldNames(tg: TypeGraph, typeIdx: number) {
...
+    if (entry.type === "directory") {
+      const resDirPath = path.join(baseDir, entry.fileName);
+      await ensureDir(resDirPath);
+      continue;
+    }
+    const resFilePath = path.join(baseDir, entry.fileName);
+    await ensureFile(resFilePath);
+    const file = await Deno.open(resFilePath, { write: true });
+    await conversion.copy(entry, file);
+    file.close();
```
zifeo, line 257: needs to be inside a try {} catch {} finally {here}

typegate/src/utils.ts
```diff
@@ -214,3 +225,82 @@ export function collectFieldNames(tg: TypeGraph, typeIdx: number) {
   }
   return { title: typ?.title, fields: [] };
 }
+
+/**
+ * base64 decode and untar at cwd/{dir}
+ */
+export async function uncompress(dir: string, tarb64: string) {
```
zifeo, line 232: suggestion, as you compress in rust, maybe it is interesting to uncompress in rust so we are sure this is symmetric?

typegate/src/utils.ts
```diff
@@ -214,3 +225,82 @@ export function collectFieldNames(tg: TypeGraph, typeIdx: number) {
...
+    const file = await Deno.open(resFilePath, { write: true });
+    await conversion.copy(entry, file);
+    file.close();
+  }
+  return baseDir;
+}
+
+export async function digestPath(path: string) {
+  const data = new TextEncoder().encode(path);
+  const buffer = await crypto.subtle.digest("SHA-1", data);
```
zifeo, line 264: there should be a hash function inside `crypto.ts` you can reuse

typegate/src/utils.ts
```diff
@@ -214,3 +225,82 @@ export function collectFieldNames(tg: TypeGraph, typeIdx: number) {
...
+    throw Error(`${filePrefix} prefix not specified`);
+  }
+  if (!tgPrefix.startsWith(tgPrefix)) {
+    throw Error(`${tgPrefix} prefix not specified`);
+  }
+  if (!base64Str.startsWith(b64Prefix)) {
+    throw Error(`${b64Prefix} prefix not specified`);
+  }
+
+  const userMainFile = fileStr.substring(filePrefix.length);
```
zifeo, line 288: can you provide a comment/example for those lines?

typegraph/python/typegraph/runtimes/deno.py
```diff
@@ -134,12 +134,15 @@ def __attrs_post_init__(self):
...
-            if os.environ.get("DONT_READ_EXTERNAL_TS_FILES"):
-                object.__setattr__(self, "code", f"file:{path}")
-            else:
-                with open(path) as f:
-                    object.__setattr__(self, "code", f.read())
+            rel_path = os.path.join("scripts", "deno", self.file)
+            path = get_absolute_path(rel_path)
+            object.__setattr__(self, "code", f"file:{path}")
+
+            # if os.environ.get("DONT_READ_EXTERNAL_TS_FILES"):
```
zifeo, line 141: @Natoandro can we remove this everywhere?

libs/common/src/archive.rs
afmika: It does not have is_empty, count or similar method. Can only add files/folder.

typegate/src/utils.ts
afmika: `untar.extract()` works like an iterator so it will have the same effect, both requiring a loop if the root has mutliple entries.

typegate/src/utils.ts
afmika: Wouldn't that mean that we have to expose the lib function to deno ? I cannot see any other way.

typegate/src/utils.ts
afmika: It encodes to base64 (has special characters)

typegraph/python/typegraph/runtimes/deno.py
Natoandro: Yes.

meta-cli/src/typegraph/postprocess.rs
```diff
@@ -111,7 +155,9 @@ pub mod deno_rt {
                 let mut mat_data: ModuleMatData = object_from_map(std::mem::take(&mut mat.data))?;
                 if !mat_data.code.starts_with("file:") {
```
Natoandro, line 156: I think this will be always false now... So the block will always be skipped...

meta-cli/src/typegraph/postprocess.rs
```diff
@@ -111,7 +155,9 @@ pub mod deno_rt {
...
                     // TODO check imported functions exist
-                    mat_data.code = transform_module(mat_data.code)?;
+                    let path = mat_data.code.strip_prefix("file:").unwrap();
```
Natoandro, line 158: Won't this always fail if ever executed??

meta-cli/src/typegraph/postprocess.rs
Natoandro: We can remove it now.

typegate/src/utils.ts
```diff
@@ -214,3 +207,74 @@ export function collectFieldNames(tg: TypeGraph, typeIdx: number) {
...
+    } catch (e) {
+      throw e;
+    } finally {
+      file?.close();
+    }
+  }
+  return baseDir;
+}
+
+export async function structureRepr(str: string): Promise<FolderRepr> {
```
Natoandro, line 251: The function name lacks of context, though the function is placed in the common `utils.ts` file...

Natoandro (commented):
Good :+1: 

Some comments and one question:

How do we handle multiple module imports in the same typegraph: won't the materializers all have the same archive??

meta-cli/src/typegraph/postprocess.rs
afmika: I do not understand, the if statement above guarantees the prefix is still present so unwrap should never fail.

libs/common/src/archive.rs
```diff
@@ -35,3 +36,41 @@ pub fn archive<P: AsRef<Path>>(folder: P) -> Result<Option<String>> {
         Ok(Some(STANDARD.encode(bytes)))
     }
 }
+
+pub fn archive_entries(dir_walker: Walk, suffix: Option<&Path>) -> Result<Option<String>> {
+    let encoder = GzEncoder::new(Vec::new(), Compression::default());
+    let mut tar = tar::Builder::new(encoder);
+
+    // skip non-relevant fs metadata
+    tar.mode(tar::HeaderMode::Deterministic);
```
zifeo, line 45: does this ignore also file like [.DS_Store](https://en.wikipedia.org/wiki/.DS_Store)?

meta-cli/src/typegraph/postprocess.rs
```diff
@@ -99,23 +103,50 @@ pub mod deno_rt {
...
                 }
-                mat.data = map_from_object(mat_data)?;
             }
-            _ => {}
+            None => "".to_string(),
+        };
+
+        let dir_walker = WalkBuilder::new(&base_rel_path)
+            .standard_filters(true)
+            .sort_by_file_path(|a, b| a.cmp(b))
```
zifeo, line 124: nice!

meta-cli/src/typegraph/postprocess.rs
```diff
@@ -99,23 +103,50 @@ pub mod deno_rt {
...
+            .unwrap()
+            .to_owned();
+
+        format!("file:{};base64:{}", file, enc_content)
+    }
+
+    fn reformat_materializer_script(mat: &mut Materializer) -> Result<()> {
+        if mat.name.as_str() == "function" {
+            let mut mat_data: FunctionMatData = object_from_map(std::mem::take(&mut mat.data))?;
+            // TODO check variable `_my_lambda` exists and is a function expression/lambda
```
zifeo, line 147: todo before merge?

typegate/deno.json
```diff
@@ -34,6 +34,8 @@
     "levenshtein": "https://deno.land/x/levenshtein@v1.0.1/mod.ts",
     "aws-sdk/client-s3": "https://esm.sh/@aws-sdk/client-s3@3.335.0?pin=v124",
     "aws-sdk/s3-request-presigner": "https://esm.sh/@aws-sdk/s3-request-presigner@3.335.0?pin=v124",
-    "outdent": "https://deno.land/x/outdent@v0.8.0/mod.ts"
+    "outdent": "https://deno.land/x/outdent@v0.8.0/mod.ts",
+    "tar": "https://deno.land/std@0.128.0/archive/tar.ts",
```
zifeo, line 38: @0.192.0

typegate/src/log.ts
```diff
@@ -4,8 +4,23 @@
 import { handlers, LevelName, Logger } from "std/log/mod.ts";
 import { basename, extname, fromFileUrl } from "std/path/mod.ts";
 import { z } from "zod";
+import { deepMerge } from "std/collections/deep_merge.ts";
 
-import { configOrExit } from "./utils.ts";
+export const configOrExit = async <T extends z.ZodRawShape>(
```
zifeo, line 9: better location indeed

typegate/src/runtimes/deno/deno.ts
```diff
@@ -103,6 +120,9 @@ export class DenoRuntime extends Runtime {
       }
     }
 
+    const tgSCriptFolder = path.join("tmp", "scripts", typegraphName);
+    (args.permissions as Deno.PermissionOptionsObject).read = [tgSCriptFolder];
```
zifeo, line 124: I would put that line below :
```
{...args.permissions, read: [...]} as Deno.X
```

typegate/src/runtimes/deno/deno_messenger.ts
```diff
@@ -28,10 +28,10 @@ export class DenoMessenger extends LazyAsyncMessenger<Worker, Task, unknown> {
               hrtime: false,
               net: true,
               // on request permissions
+              // read: ["tmp"], // default read permission
```
zifeo, line 31: let's put it `read: false` so it is disallowed if not overridden and we are safe if we forget it

typegate/src/utils.ts
zifeo: @afmika yes, you could create a new crates in libs and use it both in meta-cli and typegate native

meta-cli/src/typegraph/postprocess.rs
afmika: This comment was there before, I just removed the unused code then replaced the `match` to a `if`. Good to have indeed, but I assume it wouldn't be that trivial to do properly.

libs/common/src/archive.rs
afmika: No I don't think so. But they are usually added to gitignore.

meta-cli/src/typegraph/postprocess.rs
Natoandro: The condition is if not starting with `file:`...

zifeo (dismissed):

typegraph/core/tests/snapshots/test_sdks__typegraphs__simple.snap
```diff
@@ -219,7 +219,7 @@ input_file: typegraph/core/tests/typegraphs/simple.py
           "idempotent": true
         },
         "data": {
-          "code": "file:/three.ts;base64:H4sIAAAAAAAA/+2SwUrEMBCGc85TzFGhrGnTbKGKd6/iC6Q1q0VMlnSCCyXvbpKDloIISgvL5rtMJjOZTPLP7oasDgs0QiQbWNq0Llm9L1nFKp72G14JIsgGuBGlBSD/Zfm4MwFHsjp/0V/wJuu/AfhqldqtOwXxP/Z1/aP+vOIL/cuQH+aFbMCF608HjcoeZK/gKY7Cgz46hIkCdC1o994pexsc+e14StXpaCzCwekeB6MhDdHVBLKADnw7q1SEsxP0Jlxywhj6clp4VL2xz3cj2kG/FOD0mzYf+h48vW5TB3Lewawdn6JWobM6FAxprIhxBj62RzKZTCbzO58H2mKzAAwAAA=="
+          "code": "tmp/unpacked,tmp/unpacked/three.ts"
```
zifeo, line 222: maybe snapshot the content of the file instead of the filename?

typegraph/core/tests/test_sdks.rs
```diff
@@ -47,6 +49,45 @@ fn get_test_dir() -> PathBuf {
         .into()
 }
 
+fn replace_b64_to_entries(tg: String) -> String {
+    let mut arr: Value = serde_json::from_str(&tg).unwrap();
```
Natoandro, line 53: Maybe `Vec<Typegraph>` instead of `Value`?

zifeo (approved):

PR #346, opened 2023-06-27
feat: add grpc runtime

msga-mmm:


typegate/tests/runtimes/grpc/grpc.py
```diff
@@ -0,0 +1,68 @@
...
+    # gRPC server endpoint
+    grpc = GrpcRuntime("http://localhost:4770")
+    public = policies.public()
+
+    g.expose(
+        greet=grpc.call_method(
+            # protocol buffer file with the definition for the method to call
+            this_dir.joinpath("grpc_server/proto/helloworld.proto"),
+            # the method to call: `SayHello`
+            "/helloworld.Greeter/SayHello",
```
zifeo, line 18: has the `/` prefix any logic or common usage?

zifeo (changes requested):
Very nice. You will have the coverage once the tests are fixed (working on it now).
I added a few comments, and especially 2 important points:
1. `import "other.proto";` can you add a test with a proto import?
2. this will work for proto files accessible, but sometimes proto reflection can be used. Do you believe this can be easily support (e.g. using another materializer) ?

Natoandro:
Stale, duplicate of #819.

PR #350, opened 2023-06-28
feat: improve SDKs and rest support

zifeo:


typegate/tests/runtimes/rest/rest.py
Natoandro, line 1: I think we should rename this into `http` because the runtime name is actually `HTTPRuntime`.

Natoandro (approved):
:+1:

typegate/tests/runtimes/rest/rest.py
zifeo: Totally!

Natoandro (approved):

typegraph/core/src/lib.rs
```diff
@@ -28,16 +28,48 @@ pub mod wit {
...
+            Ok(())
+        }
+    }
 }
 
 pub struct Lib {}
 
 impl wit::core::Core for Lib {
     fn init_typegraph(params: TypegraphInitParams) -> Result<()> {
+        /*
```
Natoandro, line 64: Uh?

PR #358, opened 2023-07-06
fix(Deno): generate tarball at tg level

afmika:


meta-cli/src/typegraph/postprocess.rs
```diff
@@ -103,43 +103,53 @@ pub mod deno_rt {
...
+        }
+
+        if tg_path.is_relative() {
+            bail!(
+                "typegraph path {:?} is relative, absolute expected",
+                tg_path.display()
+            );
+        }
+
+        let tg_root = if let Some(parent) = tg_path.parent() {
```
zifeo, line 124: .context() is cleaner

typegraph/core/tests/snapshots/test_sdks__typegraphs__simple.snap
```diff
@@ -219,7 +219,7 @@ input_file: typegraph/core/tests/typegraphs/simple.ts
           "idempotent": true
         },
         "data": {
-          "code": "tmp/unpacked,tmp/unpacked/three.ts"
+          "code": "tmp/unpacked,tmp/unpacked/scripts,tmp/unpacked/scripts/deno,tmp/unpacked/scripts/deno/three.ts,tmp/unpacked/simple.py,tmp/unpacked/simple.ts"
```
zifeo, line 222: tmp are ignored by the glob, this is why the test does not pass

zifeo (commented):
Merged and fix in https://github.com/metatypedev/metatype/pull/350

typegraph/python/typegraph/runtimes/deno.py
```diff
@@ -132,10 +131,17 @@ def __attrs_post_init__(self):
             if self.code is not None:
                 raise Exception("you must only give either source file or source code")
 
-            from typegraph.graph.typegraph import get_absolute_path
+            # from typegraph.graph.typegraph import get_absolute_path
+
+            # depth 0: typegraph.py (definition)
```
zifeo, line 136: not always true, for instance if an importer is used into another file?

afmika: Definitely, this is why I dropped the inspect approach.

PR #365, opened 2023-07-17
feat(gate): generate openapi over rest endpoints

afmika:
This solves MET-205


typegate/src/engine.ts
```diff
@@ -262,7 +266,16 @@ export class Engine {
...
-      this.rest[effectToMethod[effect!]][name] = [plan, checkVariables];
+      const getVariableNames = () => {
+        const defs = unwrappedOperation.variableDefinitions ?? [];
+        return defs.map((v) => v.variable.name.value);
+      };
+
+      this.rest[effectToMethod[effect!]][name] = [
+        plan,
+        checkVariables,
+        getVariableNames,
```
zifeo, line 277: Let's store an object and pre-compute the variables name directly here.

typegate/src/services/rest_service.ts
```diff
@@ -24,6 +31,17 @@ export async function handleRest(
     const name = url.pathname.split("/").slice(
       3,
     ).join("/");
+
+    if (name == "__schema") {
```
zifeo, line 35: can you check that we have this in GET only ?

typegate/src/services/rest_service.ts
```diff
@@ -24,6 +31,17 @@ export async function handleRest(
...
+      logger.info(`rest: ${name} fetch openapi schema`);
+      const res = buildOpenAPISpecFrom(url.origin, engine);
+      headers.set("Content-Type", "application/json");
+      return new Response(res, {
+        status: 200,
+        headers,
+      });
+    }
+
```
zifeo, line 44: for GET /, let's return them something similar to [this](https://github.com/metatypedev/metatype/blob/main/typegate/src/services/playground_service.ts) with [that](https://redocly.com/docs/redoc/deployment/html/#step-2---reference-the-redoc-script)

typegate/src/services/rest_service.ts
```diff
@@ -95,3 +113,120 @@ export async function handleRest(
...
+  for (const method of Object.keys(engine.rest)) {
+    const queries = engine.rest[method];
+    for (const name of Object.keys(queries)) {
+      const [, , getVariables] = queries[name];
+
+      const parameters = getVariables().map((v) => ({
+        name: v,
+        in: "query",
+        required: true,
+        schema: { type: "string" },
```
zifeo, line 185: They can also be more complex like an object with field of int or nested objects. We may want to have everything pre-cooked at typegraph push and re-use some of the typing logic we have for the graphql request.

typegate/src/services/rest_service.ts
```diff
@@ -95,3 +113,120 @@ export async function handleRest(
...
+            500: genericErrorResponse(name, 500),
+            403: genericErrorResponse(name, 403),
+            400: genericErrorResponse(name, 400),
+          },
+        },
+      };
+    }
+  }
+
+  spec.paths[`/${title}/rest/__rest`] = {
```
zifeo, line 211: I believe you can remove that one (as in graphql the introspection does not explain itself)

zifeo (commented):
Looks good so far!

typegate/src/services/rest_service.ts
```diff
@@ -28,9 +30,11 @@ export async function handleRest(
     const queries = engine.rest[req.method];
     const url = new URL(req.url);
 
-    const name = url.pathname.split("/").slice(
+    const parts = url.pathname.split("/");
```
zifeo, line 33: const [typegraphName, ...parts] =

typegate/src/services/rest_service.ts
```diff
@@ -48,6 +52,12 @@ export async function handleRest(
       }
     }
 
+    if (req.method === "GET" && config.debug && (name === "/" || !name)) {
```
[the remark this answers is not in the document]
afmika: I check for the schema name first then the GET method, but you are right, I should refactor it a bit. 👍

zifeo (approved):

PR #375, opened 2023-07-27
feat(gate): deno worker should timeout internal functions to avoid infinite loop or similar

afmika:
Attempt solving MET-120.

There seems to be an issue when using `setTimeout` and `setInterval` in tests even if they are 'properly' cleared. (some might be cancelled prematurely?)
```
Deno runtime ... should work with async function => ./typegate/tests/utils/test.ts:148:30
error: Leaking async ops:
  - 1 async operation to sleep for a duration was started before this test, but was completed during the test. Async operations should not complete in a test if they were not started in that test.
            This is often caused by not cancelling a `setTimeout` or `setInterval` call.
```

I also tried using an external sleep source like `$ sleep 1s` but it seems to have similar issues
```
error: Leaking resources:
  - A child process stdout (rid 158) was opened before the test started, but was closed during the test. Do not close resources in a test that were not created during that test.
```

[Update]
Disable `sanitizeOps` while making sure all resources are closed properly.
```typescript
Meta.test("test title", async (t) => {
  // test body
}, { sanitizeOps: false });
```


zifeo:
@afmika Can you push the version with const inter = setInterval() and clearInterval(inter)? This should work but we may need to setup a cleaning call from the runtime shutdown hook.

typegate/src/runtimes/patterns/messenger/async_messenger.ts
```diff
@@ -34,12 +42,48 @@ export class AsyncMessenger<Broker, M, A> {
...
+            this.receive({
+              id: item.message.id,
+              error: `${maxDurationMs / 1000}s timeout exceeded (+${
+                (delta - maxDurationMs) / 1000
+              }s)`,
+            });
+            this.#pendingOperations.pop(); // O(log N)
+
+            if (config.timer_destroy_ressources) {
+              this.#stop(this.broker); // force abort
```
zifeo, line 81: It should not only stop & cancel all ongoing promises (use a specific errors so the resolver throw an error explained it has been interrupted by another one), but also restart (if lazy easy, if sync a bit more complex) and make sure next requests are successfully fulfilled

typegate/src/runtimes/patterns/messenger/async_messenger.ts
```diff
@@ -34,12 +42,48 @@ export class AsyncMessenger<Broker, M, A> {
...
+
+  initTimer() {
+    if (this.#timer === undefined) {
+      this.#timer = setInterval(() => {
+        const currentDate = Date.now();
+        const item = this.#pendingOperations.peek(); // O(1)
+        if (item !== undefined) {
+          // safe removal
+          const delta = currentDate - item.date;
+          if (this.#doneIds.has(item.message.id)) {
```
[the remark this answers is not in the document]
afmika: `pendingOperations`: automatic sort by date, the min access is constant. But scanning is still linear.
`doneIds`: scans are constants.
`doneIds` is unpopulated accordinly as the timer ticks.

typegate/src/runtimes/patterns/messenger/async_messenger.ts
afmika: `restart` as in (1) restarting the task (restart broker + re-run task) or (2) restart the broker ?

typegate/src/runtimes/patterns/messenger/async_messenger.ts
```diff
@@ -34,12 +42,48 @@ export class AsyncMessenger<Broker, M, A> {
...
+              id: item.message.id,
+              error: `${maxDurationMs / 1000}s timeout exceeded (+${
+                (delta - maxDurationMs) / 1000
+              }s)`,
+            });
+            this.#pendingOperations.pop(); // O(log N)
+
+            if (config.timer_destroy_ressources) {
+              this.#stop(this.broker); // force abort
+            } // else: let the process owner destroy the ressources
```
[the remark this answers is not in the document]
zifeo: Ok, so process owner = instance owner?

typegate/src/runtimes/patterns/messenger/async_messenger.ts
zifeo: 2 :)

typegate/src/runtimes/patterns/messenger/async_messenger.ts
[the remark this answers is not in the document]
afmika: yes! this is a hack for the infinite loop test. Clearly not the best option, but I am still working on it.

typegate/src/config.ts
```diff
@@ -39,6 +39,8 @@ const schema = {
     }
     return bytes;
   }),
+  timer_max_timeout_ms: z.number().positive().max(60000), // 1min
+  timer_destroy_ressources: z.boolean().default(true),
```
zifeo, line 43: Let's have all default at the same place: `.default(true)` shall already be taken care by line 86

typegate/tests/runtimes/deno/deno_test.ts
```diff
@@ -155,3 +156,36 @@ Meta.test("Deno runtime: reloading", async (t) => {
...
+      await gql`
+          query {
+            infiniteLoop(enable: true)
+          }
+        `
+        .expectErrorContains("timeout exceeded")
+        .on(e);
+    },
+  );
+  config.timer_destroy_ressources = configValue;
```
zifeo, line 190: Test that request can still continue afterwards?

zifeo (commented):
And same tests for Python?

typegate/tests/runtimes/deno/deno_test.ts
afmika: This was an `async ops` hack. This test seems to break the ci anyways, it makes test-typegate running indefinitely  (vs local). Still working on how to tackle it properly.

zifeo (dismissed):

typegate/src/config.ts
```diff
@@ -39,6 +39,9 @@ const schema = {
     }
     return bytes;
   }),
+  timer_max_timeout_ms: z.number().positive().max(60000),
+  timer_max_timeout_test_ms: z.number().positive().max(60000),
```
zifeo, line 43: I was more thinking on injection config.timer_max_timeout_ms in utils/mod.ts because adding an env var just for test sounds a bit heavy :)

typegate/src/runtimes/python_wasi/python_wasm_messenger.ts
```diff
@@ -14,7 +14,7 @@ export class PythonWasmMessenger extends AsyncMessenger<
   private constructor() {
     const vmMap = new Map<string, PythonVirtualMachine>();
     super(
-      vmMap,
+      () => vmMap,
```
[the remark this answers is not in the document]
afmika: In the new start scheme introduced, after stopping a broker, we call `start(broker)`.  For python runtime's case, start has no effect since all VMs are initialized at init.

zifeo (approved):

PR #384, opened 2023-08-03
feat: more stable run by fixing little bugs + typegraph upgrade

zifeo:


typegate/tests/runtimes/python_wasi/python_wasi_test.ts
```diff
@@ -117,6 +117,22 @@ Meta.test("Python WASI runtime", async (t) => {
       .on(e);
   });
 
+  await t.should("support nested reeturn", async () => {
+    await gql`
+        query {
+          hello (world :"test"){
+            test
+          }
```
afmika, line 125: Isn't this redundant? This is already tested on `return same object` test.

afmika (commented):
Nice

PR #403, opened 2023-08-24
feat(sdk): implement injection

afmika:


typegraph/core/wit/typegraph.wit
```diff
@@ -15,12 +15,14 @@ interface core {
     finalize-typegraph: func() -> result<string, error>
 
     type type-id = u32
-
     record type-base {
         name: option<string>,
+        // json
+        injection: option<string>,
```
zifeo, line 21: can we type with an enum the different injection?

typegraph/deno/src/effects.ts
```diff
@@ -23,3 +23,9 @@ export function delete_(idempotent = true): EffectDelete {
 export function update(idempotent = true): EffectUpdate {
   return { tag: "update", val: idempotent };
 }
+
+export const effectPrefix = "__effect_";
```
zifeo, line 27: I believe we shall support simple string as well so I would avoid any prefix

typegraph/python_next/typegraph_next/t.py
```diff
@@ -77,6 +78,66 @@ def optional(self, default_value: Optional[str] = None) -> "optional":
...
+                    isinstance(k, EffectType) for k in value.keys()
+                ):
+                    correct_value = {}
+                    for k, v in value.items():
+                        if not isinstance(v, str):
+                            raise Exception(
+                                f"value for field {k.name} must be a string"
+                            )
+                        correct_value[k] = proxy(v).id
+                else:
```
zifeo, line 125: same here, heavily nested

typegraph/core/wit/typegraph.wit
afmika: wit enums cannot have type payloads like in rust.

typegraph/deno/src/effects.ts
afmika: If we allow simple string then we have to reserve `update`, `delete`, `create`, and `none` for the typescript frontend. 
We can however have a workaround by making `PerEffect` explicit
```typescript
struct({ a: t.string() })
.set({ [NONE]: {a: 1}, [CREATE]: {a: 2} })  // prefix is used to distinguish between an object and a per effect object
```
becomes
```typescript
struct({ a: t.string() })
.set({ none: {a: 1}, create: {a: 2} }, "PerEffect")
```

Natoandro: We can use [symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol).

typegraph/core/wit/typegraph.wit
Natoandro: But WIT have `variant`, see [WIT types](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#wit-types).

typegraph/core/wit/typegraph.wit
```diff
@@ -15,12 +15,14 @@ interface core {
...
         // string => json string
         runtime-config: option<list<tuple<string, string>>>
     }
+    update-type-injection: func(id: u32, value: string) -> result<_, error>
```
Natoandro, line 25: I would rather not modify the type, as the same type might be used elsewhere without the injection. See the approach I used for policies.

```
record type-with-injection {
   tpe: type-id,
   injection: ...
}
```

Natoandro (changes requested):
Some suggestions.

typegraph/deno/src/utils/func_utils.ts
```diff
@@ -1,6 +1,63 @@
...
+    value !== null
+  ) {
+    // Note:
+    // Symbol changes the behavior of keys, values, entries => props are skipped
+    const symbols = [UPDATE, DELETE, CREATE, NONE];
+    const noOtherType = Object.keys(value).length == 0;
+    const isPerEffect = noOtherType &&
+      symbols
+        .map((symbol) => (value as any)?.[symbol] !== undefined)
+        .reduce(
```
Natoandro, line 32: [Array.some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

Natoandro, line 32: [`Array.some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) ?

Natoandro (approved):
Nice :+1:

Some small suggestion.

Natoandro (approved):
Nice :+1:

Some small suggestion.

typegraph/deno/src/utils/func_utils.ts
afmika: Indeed this should make it shorter/cleaner.

zifeo (approved):

PR #410, opened 2023-09-04
feat(sdk): apply syntax

afmika:


typegraph/core/src/global_store.rs
```diff
@@ -64,12 +64,62 @@ impl Store {
...
+    ) -> Result<(&Type, TypeId), TgError> {
+        let mut ret = (self.get_type(struct_id)?, struct_id);
+
+        let mut curr_path = vec![];
+        for chunk in path {
+            let unwrapped_id = self.resolve_quantifier(ret.1)?;
+            let unwrapped = self.get_type(unwrapped_id)?;
+            if let Type::Struct(t) = unwrapped {
+                let mut found = false;
+                for (k, v) in t.data.props.iter() {
```
zifeo, line 105: .find?

typegraph/core/src/typegraph.rs
```diff
@@ -334,4 +334,14 @@ impl TypegraphContext {
     pub fn find_type_index_by_store_id(&self, id: &u32) -> Option<u32> {
         self.mapping.types.get(id).copied()
     }
+
+    pub fn get_correct_id(&self, id: u32) -> Result<u32> {
+        with_store(|s| {
+            let id = s.resolve_proxy(id)?;
+            if let Some(index) = self.find_type_index_by_store_id(&id) {
```
zifeo, line 341: if there is a return or we have to handle the 2 cases (Some, None) match syntax is easier to understand

typegraph/core/src/utils/apply.rs
```diff
@@ -0,0 +1,189 @@
...
+pub fn flatten_to_sorted_items_array(path_tree: &PathTree) -> Result<Vec<ItemNode>, String> {
+    let mut index = 0;
+    let mut levels = vec![vec![ItemNode {
+        parent_index: None,
+        index,
+        node: path_tree.clone(),
+    }]];
+
+    loop {
+        if let Some(parents) = levels.last() {
```
zifeo, line 127: e.g. here match is easier to read

typegraph/core/src/utils/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+};
+
+mod apply;
+
+fn find_missing_props(
+    old_props: &Vec<(String, u32)>,
+    new_props: &Vec<(String, u32)>,
+) -> Vec<(String, u32)> {
+    let mut missing_props = vec![];
+    for (k_old, v_old) in old_props {
```
zifeo, line 20: use hashset from keys and set operation?

typegraph/deno/src/types.ts
```diff
@@ -503,6 +506,22 @@ export class Func<
...
+  apply(value: Record<string, unknown | InheritDef>) {
+    const data: Apply = { paths: [] };
+    buildApplyData(value, data);
+    const exampleId = wit_utils.genApplyb(
+      this.inp._id,
+      data,
+    );
+
+    return func(
+      // TODO: copy props?
```
zifeo, line 519: should not be necessary as immutable

typegraph/deno/src/utils/func_utils.ts
```diff
@@ -59,3 +61,52 @@ export function serializeRecordValues<T>(
 ): Array<[string, string]> {
   return Object.entries(obj).map(([k, v]) => [k, JSON.stringify(v)]);
 }
+
+export function buildApplyData(
```
zifeo, line 65: can we avoid side effect and return the modified data instead?

typegraph/core/src/utils/apply.rs
```diff
@@ -0,0 +1,189 @@
...
+            }
+            levels.push(current);
+        } else {
+            panic!("first level must be populated");
+        }
+    }
+
+    // flatten the tree to a 1D array
+    let final_size = (index + 1) as usize;
+    let mut tmp_list = vec![None; final_size];
```
zifeo, line 151: better name?

typegate/tests/typecheck/apply_syntax_test.ts
```diff
@@ -0,0 +1,94 @@
...
+  );
+
+  await t.should(
+    "compose apply and work with partial static injections",
+    async () => {
+      await gql`
+        query {
+          applyComposition (
+            student: {
+              name: "Jake"
```
zifeo, line 48: add a case with from_context()?

typegraph/core/src/utils/mod.rs
afmika: We only have to test against the names. That could work if the pairs (name, id) in the two vecs (old and new) always overlap. In this case, ids can differ even if the names are the same.

Natoandro (dismissed):

zifeo (approved):

zifeo (approved):

PR #414, opened 2023-09-21
Make `with_store` and `with_store_mut` private in `global_store` module

Natoandro:


zifeo (approved):
CleanShot 2023-09-22 at 14 49 47@2x
Congrats for simplifying!

PR #419, opened 2023-09-26
feat(sdk): custom query exec for prisma runtime

afmika:


typegate/tests/runtimes/prisma/full_prisma_mapping.py
```diff
@@ -79,4 +80,28 @@ def prisma(g: Graph):
                 "where": {"id": 10007},
             }
         ),
+        testExecuteRaw=db.execute(
+            'UPDATE "Post" SET title = $1 WHERE title like $2',
```
zifeo, line 84: I believe named arg are worth supporting to avoid ordering issue `${test}`, then you can delegate the replacement by other characters to the rust and then the final replacement to Prisma

meta-cli/src/typegraph/loader/discovery.rs
```diff
@@ -108,8 +108,8 @@ pub struct FileFilter {
 impl FileFilter {
     fn python_filters(config: &TypegraphLoaderConfig) -> Result<SpecificFilters> {
         // soon
-        // let matcher = RegexMatcher::new_line_matcher("with\\s+typegraph")?;
-        let matcher = RegexMatcher::new_line_matcher("with\\s+[Tt]ype[Gg]raph")?;
+        // let matcher = RegexMatcher::new_line_matcher("@typegraph")?;
+        let matcher = RegexMatcher::new_line_matcher("with\\s+[Tt]ype[Gg]raph|@typegraph")?;
```
zifeo, line 112: I believe you can include the parenthesis `@typegraph(`

typegate/src/runtimes/prisma/prisma.ts
```diff
@@ -152,6 +152,7 @@ export class PrismaRuntime {
         throw new ResolverError(res.errors[0].user_facing_error.message);
       }
 
+      // console.log("prisma result", res);
```
zifeo, line 155: rm?

typegate/tests/auto/__snapshots__/auto_test.ts.snap
```diff
@@ -365,15 +365,9 @@ snapshot[`Auto-tests for prisma 1`] = `
...
-    {
-      extensions: {},
-      locations: [],
-      message: "Validation errors:
-  - at <value>.where.firstname: Value does not match to any variant of the either type
-",
-      path: [],
-    },
-  ],
+  data: {
```
zifeo, line 368: 🔥

typegate/tests/runtimes/prisma/full_prisma_mapping.py
```diff
@@ -79,4 +80,38 @@ def prisma(g: Graph):
...
+                }
+            ),
+            EffectUpdate(True),
+        ),
+        testQueryRaw=db.query_raw(
+            """
+                WITH tmp AS (
+                    SELECT id, title, (views + likes) as reactions FROM "Post"
+                    WHERE title LIKE ${title}
+                ) SELECT * FROM tmp WHERE id IN (${one}, ${two})
```
zifeo, line 98: Great! I believe there is one more use case with the `select * from a WHERE b IN (${list})` where `t.array(t.string())`. At the list cardinality could be varying we need to support any size.

typegraph/python/typegraph/providers/prisma/runtimes/prisma.py
```diff
@@ -150,28 +150,28 @@ def link(
...
-                {
-                    "parameters": t.json().optional().default("[]"),
-                }
-            ).named(f"ExecuteRawInp_{TypegraphContext.get_active().next_type_id()}"),
+            inp,
             t.integer(),
             PrismaOperationMat(self, query, "executeRaw", effect=effect),
         )
 
+    def as_column(tpe: t.TypeNode) -> t.struct:
```
zifeo, line 167: can we move this to the typegate so the typing on the typegraph is transparent to the user?

zifeo (approved):

PR #429, opened 2023-10-03
chore: Migrate doc typegraphs

Natoandro:


typegraph/python_next/typegraph_next/runtimes/random.py
```diff
@@ -18,7 +18,7 @@
 
 
 class RandomRuntime(Runtime):
-    def __init__(self, seed: int, reset: Optional[str] = ""):
+    def __init__(self, seed: int = 0, reset: Optional[str] = ""):
```
zifeo, line 21: Most of the lib use a random seed if not set, I would use the same here

website/docs/tutorials/adding-more-runtimes/prisma-no-sugar.py
```diff
@@ -1,28 +1,30 @@
...
-            effect=effects.create(),
-        ),
-    )
+# # skip:start
+# from typegraph_next import typegraph, effects, t, Graph
+# from typegraph_next.providers.prisma import PrismaRuntime
+#
+#
+# @typegraph()
+# def prisma_no_sugar(g: Graph):
```
zifeo, line 7: any issue with that one?

Natoandro: We do not expose materializers directly to the user anymore.

PR #431, opened 2023-10-03
fix(gate): introspection if func has the same input/output

afmika:


afmika:
> lgtm, can you add an introspection snapshot for all auto tests?



> lgtm, can you add an introspection snapshot for all auto tests?
I've already added this one. Or did I miss something? 
https://github.com/metatypedev/metatype/blob/MET-186-introspection-fails-on-http-localhost-3000-docs-guides-authentication/typegate/tests/utils/autotest.ts#L95

zifeo:
@afmika indeed 👍

PR #432, opened 2023-10-03
refactor(typegraph_core): Simplify private rust SDK

Natoandro:


typegraph/core/src/runtimes/prisma/relationship/mod.rs
```diff
@@ -162,6 +162,17 @@ impl PrismaLink {
     }
 }
 
+impl TypeBuilder for PrismaLink {
+    fn build(&self) -> Result<TypeId> {
+        self.build_link()
+    }
+}
+
+#[allow(dead_code)]
```
zifeo, line 171: why dead?

zifeo (approved):

typegraph/core/src/runtimes/prisma/relationship/mod.rs
Natoandro: It is only used in the tests...

afmika (commented):
nice 👍

PR #465, opened 2023-10-24
feat(prisma/migrations): Default value on new column

Natoandro:
- [x] Display a more concise error message for new column that failed the NON NULL constraint during the migration.
- [x] Enable user to set default value on new NON NULL column.


typegate/deno.json
```diff
@@ -36,6 +36,7 @@
     "outdent": "https://deno.land/x/outdent@v0.8.0/mod.ts",
     "json-schema-faker": "npm:json-schema-faker@0.5.3",
     "ajv": "https://esm.sh/ajv@8.12.0?pin=v131",
-    "@typegraph/sdk/": "../typegraph/deno/src/"
+    "@typegraph/sdk/": "../typegraph/deno/src/",
+    "@test-utils/": "./tests/utils/"
```
zifeo, line 40: any reason to prefix with @?

typegate/tests/utils/test_module.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  shell(cmd: string[], options: ShellOptions = {}) {
+    return shell(cmd, {
+      currentDir: this.currentDir,
+      ...options,
+    });
+  }
+
+  async cli(...args: string[]): Promise<ShellOutput>;
+  async cli(options: ShellOptions, ...args: string[]): Promise<ShellOutput>;
```
zifeo, line 23: can we remove alias signature?

typegate/tests/e2e/cli/deploy_test.ts
```diff
@@ -1,14 +1,175 @@
...
+    await t.should("load first version of the typegraph", async () => {
+      await reset();
+      await writeTypegraph(null);
+    });
+
+    await deploy();
+
+    await t.should("insert records", async () => {
+      const e = t.getTypegraphEngine(tgName);
+      if (!e) {
```
zifeo, line 140: .on(e) should any check on the existence of the typegraph, move that check there?

PR #466, opened 2023-10-25
feat: `mt_deno`

Yohe-Am:
Todo in this PR:
- [x] Expose `bench` and `test` sub commands from deno

This approach gives us a binary size of 101 MiB stripped and 143 MiB prior.


Natoandro (approved):

Cargo.toml
```diff
@@ -1,12 +1,21 @@
 [workspace]
 members = [
+  "libs/*",
```
zifeo, line 3: sweet

libs/deno/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "mt_deno"
+edition.workspace = true
```
zifeo, line 3: to change in all projects?

libs/deno/runtime.js
```diff
@@ -0,0 +1,11 @@
+// console.log("runtime.js is running!");
+
+// const { core } = Deno[Deno.internal];
+const { core } = Deno;
```
zifeo, line 4: no ts supported (I saw they are ignored?)

typegate/deno.jsonc
```diff
@@ -7,8 +7,8 @@
   "tasks": {
     "comment1": "echo cwd is by default the directory of deno.json",
     "comment2": "echo cannot restrict ffi to a lib https://github.com/denoland/deno/issues/15511",
-    "run": "cd .. && deno run --config=typegate/deno.json --unstable --allow-run=hostname --allow-sys --allow-env --allow-hrtime --allow-write=tmp --allow-ffi --allow-read=.  --allow-net typegate/src/main.ts",
-    "test": "cd .. && deno test --trace-ops --config=typegate/deno.json --unstable --allow-run=cargo,hostname,target/debug/meta,git,python3,rm,mkdir --allow-sys --allow-env --allow-hrtime --allow-write=tmp,typegate/tests --allow-ffi --allow-read=. --allow-net"
+    "run": "cd .. && deno run --config=typegate/deno.jsonc --unstable --allow-run=hostname --allow-sys --allow-env --allow-hrtime --allow-write=tmp --allow-ffi --allow-read=.  --allow-net typegate/src/main.ts",
+    "test": "cd .. && RUST_MIN_STACK=8388608 cargo x deno test --config=typegate/deno.jsonc"
```
zifeo, line 11: cannot this be set in build.rs or similar?

Yohe-Am: This one's for setting the [stack sizes of new threads](https://doc.rust-lang.org/std/thread/) i.e. a runtime variable not build time. This means that the default thread won't sport this size but this is enough for our purposes.  But yeah, I'll find a better place to put it in.

libs/deno/runtime.js
Yohe-Am: I don't think so, they're using JS in the deno repo as well. I'll look into it, we can probably add a buildstep if necessary.

libs/deno/Cargo.toml
Yohe-Am: I did start down that road but reverted after thinking it might be better to do it in another PR.

whiz.yaml
```diff
@@ -28,6 +28,17 @@ typegate2:
     <<: *tpenv
     TG_PORT: "7892"
 
+typegate3:
+  <<: *tp
+  workdir: .
+  env:
+    <<: *tpenv
+    CLICOLOR_FORCE: "1"
```
zifeo, line 36: to add in tpenv?

zifeo (commented):
Little questions but very excited to merge this!

Cargo.toml
```diff
@@ -1,12 +1,21 @@
...
 ]
 
 resolver = "2"
+
+[workspace.package]
+version = "0.2.5-0+dev"
+edition = "2021"
+
+[workspace.dependencies]
+mt_deno = { path = "libs/deno/"}
```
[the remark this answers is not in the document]
Yohe-Am: Nothing technical. Just a matter of avoiding all the different permutations of `path = ".../../my/crate` that internal crates are burdened with. I was going to propose all shared dependencies to the `workspace` section as well. A single place to specify the version of say, `clap` or `tokio`.

Cargo.toml
```diff
@@ -1,12 +1,21 @@
...
-  "libs/xtask",
-  "libs/macros",
-  "typegraph/core"
+  "typegraph/core",
 ]
 
 resolver = "2"
+
+[workspace.package]
+version = "0.2.5-0+dev"
```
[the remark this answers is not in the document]
Yohe-Am: Another thing I was going to propose was that we move all the crate manifests to `version.workspace = true`. Should obsolete the `cargo` section of `dev/lock.yaml`.

libs/deno/lib.rs
```diff
@@ -0,0 +1,299 @@
+#![allow(dead_code, clippy::let_and_return)]
```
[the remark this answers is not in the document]
Yohe-Am: Should be possible, I'll fix that and incorp all the other minor improvements you mention.

meta-cli/src/cli/typegate.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+#[async_trait]
+impl Action for Typegate {
+    async fn run(&self, _gen_args: GenArgs) -> Result<()> {
+        unreachable!()
+    }
+}
+
+pub fn command(cmd: Typegate, gen_args: GenArgs) -> Result<()> {
```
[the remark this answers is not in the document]
Yohe-Am: Incorporating it into `dev` is good idea. Is there no scenario wherein folks might just want to run the `typegate` separately?

meta-cli/src/main.rs
```diff
@@ -43,13 +45,44 @@ async fn main() -> Result<()> {
...
+    //   panic hook to implement this behaviour.
+    // - We print a message to stderr to indicate that this is a bug in Deno, and
+    //   should be reported to us.
+    let orig_hook = std::panic::take_hook();
+    std::panic::set_hook(Box::new(move |panic_info| {
+        eprintln!("\n============================================================");
+        eprintln!("Metatype has panicked. This is a bug, please report this");
+        eprintln!("at https://github.com/metatypedev/metatype/issues/new.");
+        eprintln!("If you can reliably reproduce this panic, include the");
+        eprintln!("reproduction steps and re-run with the RUST_BACKTRACE=1 env");
```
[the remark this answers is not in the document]
Yohe-Am: Good idea. I considered just running the `doctor` subcommand inside the panic hook itself but that entails calling out to the kernel for io and the shell and that doesn't seem reliable when unwinding from a panic. I've opted to add that suggestion to the text.

Cargo.toml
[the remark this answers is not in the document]
zifeo: @Yohe-Am go :)

meta-cli/src/cli/typegate.rs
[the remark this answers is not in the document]
zifeo: yes, the docker image for production

meta-cli/src/main.rs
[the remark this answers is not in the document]
zifeo: 👍

Cargo.toml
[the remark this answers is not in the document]
Yohe-Am: Maybe I should open an issue for this? Where others can propose small QOL improvements for the codebase that we'll then take care in a single PR?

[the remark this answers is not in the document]
zifeo: I would integrate those small improvements as part of the other pull request as you go :)

zifeo (approved):

meta-cli/Cargo.toml
```diff
@@ -37,7 +41,9 @@ tokio = { version = "1.33.0", features = ["full"] }
 futures = "0.3.28"
 log = "0.4.20"
 env_logger = "0.10.0"
-clap = { version = "4.4.6", features = ["derive", "env"] }
+clap = { version = "=4.3.3", features = ["derive", "env"] }
```
Natoandro, line 44: Why that specific version?

Natoandro (approved):
Great!

meta-cli/Cargo.toml
Yohe-Am: The `deno` crate pulls in that version and it triggered that edge case where cargo refuses to compile because "couldn't choose between two versions". It happily compiles duplicate versions of the same crate elsewhere, I ought to look into what causes that. But yeah, I opted to downgrade as we did not seem to be relying on anything from 4.4 and duplicates means bloat.

afmika (approved):
Very nice! 
Looking forward using this new setup.

zifeo (approved):

meta-cli/Cargo.toml
Natoandro: Could we add a comment?

PR #471, opened 2023-10-30
feat(cli): Actor model

Natoandro:


meta-cli/src/deploy/actors/console.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl Handler<Error> for ConsoleActor {
+    type Result = ();
+
+    fn handle(&mut self, msg: Error, _ctx: &mut Context<Self>) -> Self::Result {
+        log::error!("{}", msg.0);
+    }
+}
+
+macro_rules! info {
```
zifeo, line 69: Maybe here a trait would be similar? Macros may be harder to grasp for people getting familiar with the project, thus I would keep them when really necessary. WDYT?

meta-cli/src/deploy/actors/console.rs
```diff
@@ -0,0 +1,90 @@
...
+    ($addr:expr, $($arg:tt)*) => {
+        $addr.do_send($crate::deploy::actors::console::Warning(format!($($arg)*)))
+    };
+}
+
+macro_rules! error {
+    ($addr:expr, $($arg:tt)*) => {
+        $addr.do_send($crate::deploy::actors::console::Error(format!($($arg)*)))
+    };
+}
```
afmika, line 86: I think this approach hides side effects, why not use something like `addr.do_send(warning!(...))`. Using macros that way make more sense imo.

Yohe-Am (approved):

zifeo (commented):
What about the commented parts? Little fix needed for the CI :)

Yohe-Am (approved):

Yohe-Am (dismissed):

PR #485, opened 2023-11-09
fix(tests): Use  temporary directories for test git repositories

Natoandro:


Yohe-Am (approved):

typegate/tests/utils/shell.ts
```diff
@@ -61,9 +65,16 @@ export async function shell(
 
   const { code, success } = await p.status;
 
+  const out = `-- start STDOUT --\n${res.stdout}\n-- end STDOUT --`;
```
zifeo, line 68: is this temp?

zifeo (commented):
lgtm, I suppose the remaining failure is two typegraph re-using the same schema, we shall enforce this not to happen

typegate/tests/utils/test.ts
```diff
@@ -52,8 +52,13 @@ function serve(typegate: Typegate, port: number): () => void {
 
 type MetaTestCleanupFn = () => void | Promise<void>;
 
+const defaultCli = await createMetaCli(shell);
+
 export class MetaTest {
   private cleanups: MetaTestCleanupFn[] = [];
+  shell = shell;
+  cli = defaultCli;
```
zifeo, line 60: maybe t.meta is a name more coherent?

typegate/tests/utils/shell.ts
Natoandro: Yes. I'm going to clean it up.

afmika (approved):

PR #486, opened 2023-11-10
fix: small fixes

afmika:


typegraph/core/src/runtimes/prisma/type_generation/additional_filters.rs
```diff
@@ -24,7 +24,7 @@ pub struct Skip;
 
 impl TypeGen for Skip {
     fn generate(&self, _context: &PrismaContext) -> Result<TypeId> {
-        t::integer().x_min(0).named(self.name()).build()
```
zifeo, line 27: x_min(-1) or is there a non-exclusive flag?

Natoandro: `min(0)` would work.

Natoandro (approved):

zifeo (approved):

PR #487, opened 2023-11-12
feat: docs update + project structure

zifeo:


website/docs/reference/types/functions.mdx
```diff
@@ -0,0 +1,15 @@
...
+## Effects
+
+Effects are a property of materializers and help categorization what happens to data when it gets transformed. Although they are similar to REST verbs and SQL statements, there is no direct one-to-one mapping.
+
+| Effects  | REST verbs | SQL statements       |
+| -------- | ---------- | -------------------- |
+| `none`   | GET        | `SELECT`             |
+| `create` | POST       | `INSERT`             |
+| `update` | PUT/PATCH  | `UPDATE`             |
+| `upsert` | PUT        | `INSERT ON CONFLICT` |
```
Natoandro, line 12: I think we removed the `upsert` effect.

typegate/tests/e2e/templates/templates_test.ts
```diff
@@ -29,9 +29,10 @@ const modifiers: Record<string, (dir: string) => Promise<void> | void> = {
       );
     }
   },
+  "node": () => {},
```
zifeo, line 32: You have here the ability to replace the import to match to corresponding one in the project (e.g. we want to test on current commit, not the previous published version): see the function above for Deno. For node, you will have to edit package.json and make sure https://github.com/metatypedev/metatype/blob/main/dev/publish.ts is run to create the package.

dev/lock.yml
```diff
@@ -51,9 +51,9 @@ dev:
     examples/**/pyproject.toml:
       (typegraph = ").+("): METATYPE_VERSION
     examples/**/package.json:
-      '(\s*"@typegraph\/sdk"\s*:\s*"\^).+(",?)': METATYPE_VERSION
+      '(\s*"@typegraph\/sdk"\s*:\s*"\^).+(",?)': PUBLISHED_VERSION
     examples/**/*.ts:
-      '(import\s+.+\s+from "npm:@typegraph\/sdk@)[^\/]+(\/.+";)': METATYPE_VERSION
+      '(import\s+.+\s+from "npm:@typegraph\/sdk@)[^\/]+(\/.+";)': PUBLISHED_VERSION
```
zifeo, line 56: see below

zifeo (commented):
looks almost good, congrats!

Yohe-Am:
Oof. I thought the "Squash and merge" button would wait for reviews but looks like that behavior is only for protected branches. I guess I'll address any feedback on top of this branch.

Yohe-Am:
A second tutorial that focuses on the GraphQl and HTTP runtimes showcasing how easy it is to consume third-party APIs and more importantly, how to have multiple typegraphs work together....might be a good idea.

Yohe-Am:
Note: when locally building the website, you might run into strange `EPIPE` issues. It only shows up intermittently and I believe it's something introduced along the upgrade to docusaurus 3.0.0. I wasn't able to figure out what was the issue, could be something with my local env. Just wanted to give a word of warning.

That aside, this is good to go.

PR #511, opened 2023-12-06
fix(gate): pushing a typegraph must not timeout

afmika:
### Describe your change

Disable timeout on policy evaluation.

### Motivation and context

[MET-296](https://metatype.atlassian.net/jira/software/c/projects/MET/boards/2?selectedIssue=MET-296)


afmika:
> Why not use the message number to pass a longer timeout for that message? Security wise I feel we should still keep some kind of a timeout especially in the first run. Think of a forever loop in the loading of the file.

The timeout tick applies to the whole runtime not to a message specifically, but maybe we can achieve a similar result by setting a number of heartbeats per message (default 1) and decrement per tick.

zifeo:
Indeed, I would send straight after a message and kill it if no answer, this should be enough to detect a forever loop

Yohe-Am (approved):

PR #524, opened 2023-12-11
chore(deps): bump deno to 1.39.0

Yohe-Am:
### Describe your change

Brings in the latest deno version fixing the incompatibility it introduces.

This also bumps rust to `1.74.1`, needed to support latest deno.

### Motivation and context

There's a bug in 1.38.1 around something called "promise rings".

### Migration notes

No changes required by end users.

### Checklist


libs/deno/src/lib.rs
```diff
@@ -62,10 +62,10 @@ pub async fn run(
     );
 
     let flags = args::Flags {
-        subcommand: args::DenoSubcommand::Run(args::RunFlags {
-            script: main_module.path().to_owned(),
-            watch: None,
-        }),
+        // subcommand: args::DenoSubcommand::Run(args::RunFlags {
```
zifeo, line 65: weird... want to add a comment or remove it completely? is there a link to an upstream issue open on the deno tracker?

zifeo (approved):

libs/deno/src/lib.rs
Yohe-Am: I looked through the diffs on the deno repo and it seems they've done some work in the worker/worker factory section of the repo. Unable to locate the issue though (not on their bugtracker either). It also seems broken on 1.39 which I've now changed this PR to be based on.

zifeo (approved):

PR #528, opened 2023-12-19
feat(gate): redis-less mode

afmika:
### Describe your change

Fallback to `MemoryRegister` and `NoLimiter` if typegate is unable to connect to Redis.

### Motivation and context

Enable Redis-Less mode.

### Checklist


Yohe-Am (approved):

typegate/src/main.ts
```diff
@@ -46,19 +48,30 @@ init_native();
...
-  deferredTypegate,
-  redisConfig,
-);
-const limiter = await RedisRateLimiter.init(redisConfig);
-const typegate = new Typegate(register, limiter);
+let register: Register | undefined;
+let limiter: RateLimiter | undefined;
+
+try {
+  register = await ReplicatedRegister.init(deferredTypegate, redisConfig);
```
zifeo, line 55: can we me it explicit if the flag is not set? I would not love to have the typegate working because there was a connection failure, this should indeed be allowed to happen

zifeo (changes requested):

typegate/src/main.ts
zifeo: What would happen if allow_redisless_mode_on_fail is set to true and that a redis instance is provided? The same case can happen and I would argue that would be surprise for the user.

For them, the configuration should be explicit so it follows their mental model: if there is no redis instance given, then it is redisless, if there is one would connect anyway and let the process fail (no try/catch).

zifeo (changes requested):

typegate/src/main.ts
afmika: yeah I see, good point, so the idea here is to always use Redis if a config is given (and allow it to fail) otherwise, in case the config is not provided AND the flag for Redis-less mode is enabled, just use `MemoryRegister` and `NoLimiter`?

zifeo: Let's just check if redis_url is set to "none" (explicit string) that should be enough, also make sure that the typegate embedded inside the cli has this flag set

zifeo: @afmika yes, you can even take the following logic inside the cli : if the redis_url does not exists, then fallback to explicit "none" for the process launching the typegate

afmika: Fixing it.

zifeo (approved):

PR #537, opened 2024-01-01
feat: Limited parallel loads

Natoandro:
### Describe your change

Limit the number of parallel loads.
The default max is the number of CPU cores, but it can be set with the option `--max-parallel-loads=N`.

### Motivation and context

When we have a high number of typegraphs, the loader processes exhaust the CPU load.

### Migration notes

_No migration needed._

### Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,7 +4,7 @@ on:
       - v*
 
 env:
-  GHJK_VERSION: "7f63fb7"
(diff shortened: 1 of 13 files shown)


Yohe-Am (dismissed):

afmika (dismissed):
Nice! espcially knowing that posprocess may do a lot of script compression stuff in parallel.

zifeo (approved):

PR #546, opened 2024-01-12
refactor(ci): `workflow_dispatch` for release wflow

Yohe-Am:
- Manual trigger for release wflow
- Adds missing ghjk step for `vscode-extension` release job

#### Motivation and context

Provides recovery path for when things break.

#### Migration notes

Doesn't affect end users.

### Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,11 @@ on:
         required: true
         type: boolean
         default: true
+      checkBump:
+        description: check-bump adds a release entry to github so it's disabled by default.
+        required: true
+        type: boolean
+        default: false
   push:
       tags:
         - v*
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -11,36 +23,50 @@ env:
 jobs:
   check-bump:
     runs-on: ubuntu-latest
-    if: github.ref_type == 'tag'
-
+    if: github.ref_type == 'tag' 
+      || (
+        github.event_name == 'workflow_dispatch'
+        && inputs.checkBump
```
afmika, line 29: Question, how is this event triggered?

afmika (dismissed):

.github/workflows/release.yml
Yohe-Am: Workflow dispatch can be done through the github ui, in the actions tab, or their REST API.

Yohe-Am:
I thought these flakes we're seeing when deno downloads the modules was a problem with `deno.land/x` but it just flaked out with an npm package. It just might be bug in the deno resolver. Let's be upgrade to latest deno (1.39.2) and ig I'll go looking for a ticket.
image

Yohe-Am:
Strange thing is I've never seen this locally, only on the CI and I've cleaned out my DENO_DIR scores of times over the past week.

afmika (approved):

Yohe-Am:
Regarding the bug Natondoro run into [here](https://github.com/denoland/deno/issues/21866), looks like that was indeed a registry issue as discussed [here](https://github.com/denoland/deno/issues/21866).

PR #552, opened 2024-01-15
docs(website): move typegraphs in separate folder + add ts version

afmika:
#### Motivation and context

Better organization + typescript examples.

#### Migration notes

N/A

### Checklist


Yohe-Am (dismissed):
Good stuff. What do you think of moving the typegraphs to a top level `examples` dir?

zifeo:
Nice work, can you add a test (similar to the e2e ones) that checks that all ts/py typegraph serialise to the same ones? 👍 for moving into the existing example folder if there is no blocker with running whiz locally (see the related task)

Natoandro (approved):
Can you add a test that checks if they all serializes to the same JSON?

typegate/tests/e2e/website/website_test.ts
```diff
@@ -0,0 +1,101 @@
...
+async function testSerializeAllPairs(t: MetaTest, dirPath: string) {
+  const tempDir = Deno.makeTempDirSync({
+    dir: config.tmp_dir,
+  });
+
+  copySync(resolve(dirPath), tempDir, { overwrite: true });
+
+  const skip = [
+    // FIXME:
+    // "rest", // python and typescript generates different indentation for g.rest
```
zifeo, line 31: you can remove the spacing in the sdk

afmika: Sorry that was a bad diagnosis, the graphql queries was already formatted on the sdk.

Seems like most of the issues I had where around the typegraph name, type `rename` (not equivalent to the old `{ name: <name>}`), the stacktrace was just a bit tough to debug for some typegraphs.

I did have formatting issues  with the python runtime though, the python frontend parse its own source then makes some transformation to the function/lambda provided so it's never stricly equal to the typescript version unless we see what python generate first then give it to the typescript version.

Yohe-Am (approved):

zifeo (approved):

PR #555, opened 2024-01-16
feat: flattened context

Natoandro:
Flatten profile fields in the context.

So instead of
```js
{
    provider: 'github',
    accessToken: 'xxxxxxxxxxxxxxxxxxxxxxxxxx',
    refreshToken: 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
    refreshAt: 1704717676,
    profile: { id: '43663718' },
    exp: 1707280877,
    iat: 1704688876
}
```

we would have:
```js
{
    provider: 'github',
    accessToken: 'xxxxxxxxxxxxxxxxxxxxxxxxxx',
    refreshToken: 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
    refreshAt: 1704717676,
    'profile.id': '43663718',
    exp: 1707280877,
    iat: 1704688876
 }
```

#### Motivation and context

It was impossible to get the nested id into a `from_context` injection.
Now we can inject `.from_context("profile.id")`.

### Checklist

### DIFF
--- a/typegate/src/services/auth/mod.ts
+++ b/typegate/src/services/auth/mod.ts
@@ -50,14 +50,17 @@ export function initAuth(
   }
 }
 
+export type ProfileClaims = {
+  [key: `profile.${string}`]: unknown;
+};
+
 export type JWTClaims = {
   provider: string;
   accessToken: string;
   refreshToken: string;
   refreshAt: number;
-  profile: null | Record<string, unknown>;
   scope?: string[];
-};
+} & ProfileClaims;
 
 export async function ensureJWT(
   request: Request,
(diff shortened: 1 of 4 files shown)


afmika (commented):
How about enabling dot notation directly from context? I think it is best if we let the object as it is.
We split by the dots and walk the path iteratively,  this should preserve compatibility with the current context implementation. wdyt?
Ref: https://github.com/metatypedev/metatype/blob/main/typegate/src/engine/planner/args.ts#L624

zifeo:
@Natoandro adding to what @afmika is suggested, I believe this is on the app layer. I expect that the user of those auth methods write their own profiler function in deno (or else) that rewrite the context as they need and enrich it with the permissions they needed

Yohe-Am (approved):

PR #556, opened 2024-01-16
feat(prisma): Add support multi-field ids and id on struct

Natoandro:
Adds support for more advanced id fields and unique constraints:
- Multi-field id
- Id on struct (foreign key)
- Multi-field unique constraints
- Unique constraint on struct (foreign key)

#### Motivation and context

Support these kind of construct:

```python
user = t.struct(
  {
    "authProvider": t.string().from_context("provider"),
    "profileId": t.string().from_context("profile.id"),
    # ...
  },
  config={"id": ["authProvider", "profileId"]},
).rename("User")

project = t.struct(
  {
    "id": t.uuid(as_id=True, config=["auto"]),
    "owner": g.ref("Account"),
    "name": t.string(min=3, pattern="^[A-Za-z_-]$"),
  },
  config={"unique": [["owner", "name"]]}
).rename("Project")
```

#### Migration notes

_No migration needed._

### Checklist

### DIFF
--- a/typegate/tests/runtimes/prisma/full_prisma_mapping_test.ts
+++ b/typegate/tests/runtimes/prisma/full_prisma_mapping_test.ts
@@ -41,7 +41,7 @@ Meta.test("prisma full mapping", async (t) => {
                 }
               }
               extended_profile: {
-                create: { id: 10111, bio: "Some bio 1" }
+                create: { bio: "Some bio 1" }
               }
             }
           ) 
@@ -739,7 +739,6 @@ Meta.test("prisma full mapping", async (t) => {
         ) {
           id
           extended_profile {
-            id
             bio
           }
         }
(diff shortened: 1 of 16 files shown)


typegate/tests/runtimes/prisma/full_prisma_mapping_test.ts
```diff
@@ -41,7 +41,7 @@ Meta.test("prisma full mapping", async (t) => {
                 }
               }
               extended_profile: {
-                create: { id: 10111, bio: "Some bio 1" }
+                create: { bio: "Some bio 1" }
```
afmika, line 44: ` config={"id": ["user"]}` sets it auto by default or implied auto if undefined?

afmika (approved):

typegate/tests/runtimes/prisma/full_prisma_mapping_test.ts
Natoandro: The `userId` field (for the foreign key) would be the same as the `id` of the User on which it is created. It is always set by prisma for relationship fields.
In this case, it coincides with the id field.

PR #559, opened 2024-01-20
fix(cli)!: bad conditionals in loader detection

Yohe-Am:
- Fix bug in js typegraph loader detection.
- FIx bug where the cwd for loader was overridden in `loader_cmd` despite being set in `get_loader_cmd`.
- Update to latest ghjk
- Refresh `setup` task in `whiz.yaml`

#### Motivation and context

The old conditionals were faulty.

#### Migration notes

- `node` and `bun` loaders are now run with the `cwd` set to the directory of the typegraph as opposed to the `metatype.yml` file. This should resolve settings to the nearest `package.json` despite location.

### Checklist


zifeo (dismissed):

Natoandro (approved):

zifeo (approved):

examples/templates/node/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
     "compilerOptions": {
         "strict": true,
-        "moduleResolution": "node16",
-        "module": "NodeNext",
-        "esModuleInterop": true        
+        "moduleResolution": "node",
```
zifeo, line 4: I believe the oldest combat setup is:
```
"module": "ES2022",
"moduleResolution": "node16",
```

Was there any issue so we had to increase it @afmika?

zifeo (approved):

examples/templates/node/tsconfig.json
Yohe-Am: Oh, my bad. I didn't mean to commit that. I was messing around with that because the old `load-tgraph` (which you can see I removed in `examples/templates/node/packages.json`) wouldn't work for some reason I and I was trying to fix that.

zifeo (approved):

Issue #560, opened 2024-01-20
`DenoRuntime.import` is unable to locate modules when used from cli bundled typegate


Yohe-Am:
What it says on the tin can. Error is shown here:

![image](https://uploads.linear.app/8a044944-4e60-4301-b1f0-8807e48a5fca/ccb528c8-486f-4f9e-bc55-01fad8bab033/7ab9afcd-6695-4aaf-8b9c-d551f2c36228?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzhhMDQ0OTQ0LTRlNjAtNDMwMS1iMWYwLTg4MDdlNDhhNWZjYS9jY2I1MjhjOC00ODZmLTRmOWUtYmM1NS0wMWZhZDhiYWIwMzMvN2FiOWFmY2QtNjY5NS00YWFmLThiOWMtZDU1MWYyYzM2MjI4IiwiaWF0IjoxNzA1ODU0NjA5LCJleHAiOjMzMjc2NDE0NjA5fQ.jNiRyJ47K4-HtNaJvRccIghETIT0R1fI7PDhMAnSdyc "=588x")

zifeo:
Test

zifeo:
Test

zifeo:
Test

PR #564, opened 2024-01-23
fix(cli, typegate): explicitly import DenoRuntime.import modules from path

Yohe-Am:
When using the dynamic `import` function, if you provide it a raw path like `/foo/bar/baz` and deno detects the current module's loaded from a remote host, it'll convert it to a http url. Reasonable behavior but it turned out to be the cause of #560.

This pr fixes this issue along with:
- Puts contents of `main.ts` in a try/catch block for better error logging.
- Fixes minor permission bugs with the bundled runtime.

#### Motivation and context

#560 

#### Migration notes

_No changes required_

### Checklist

### DIFF
--- a/typegate/engine/src/lib.rs
+++ b/typegate/engine/src/lib.rs
@@ -107,23 +107,24 @@ pub async fn launch_typegate_deno(
         .ok_or_else(|| std::env::set_var("REDIS_URL", "none"))
         .ok();
 
+    use std::str::FromStr;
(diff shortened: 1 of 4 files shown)


Yohe-Am:
Not sure how to test this though. Should I break ground on tests for the bundled runtime?

zifeo (dismissed):

afmika (approved):

PR #566, opened 2024-01-25
feat(sdk): testing framework integration

afmika:
#### Motivation and context

Enable deploying directly from the typegraph instead of always relying on meta-cli.

#### Migration notes

N/A

### Checklist


examples/deploy/deploy.mjs
```diff
@@ -0,0 +1,34 @@
+import { Policy, t, typegraph } from "@typegraph/sdk/index.js";
+import { DenoRuntime } from "@typegraph/sdk/runtimes/deno.js";
+import { tgDeploy } from "@typegraph/sdk/tg_deploy.js";
```
zifeo, line 3: .mjs is required or can we use simple .ts or .mts?

libs/common/src/archive.rs
```diff
@@ -78,6 +79,27 @@ pub fn archive_entries(dir_walker: Walk, prefix: Option<&Path>) -> Result<Option
...
+    let encoder = GzEncoder::new(Vec::new(), Compression::default());
+    let mut tar = tar::Builder::new(encoder);
+
+    tar.mode(tar::HeaderMode::Deterministic);
+
+    for (path, bytes) in entries.iter() {
+        let mut header = Header::new_old();
+        header.set_size(bytes.len() as u64);
+        header.set_entry_type(tar::EntryType::Regular);
+        header.set_mode(0o666); // rw-rw-rw
```
zifeo, line 92: any reason for changing the permissions? maybe to include in a comment

zifeo (commented):
first rought review 👍

examples/deploy/deploy.mjs
afmika: No mjs is not required, but I guess it's the minimal setup for a node/deno typegraph without typescript.

libs/common/src/archive.rs
zifeo: 👍

typegraph/node/sdk/src/tg_deploy.ts
```diff
@@ -0,0 +1,75 @@
...
+    method: "POST",
+    headers,
+    body: JSON.stringify(gqlBody(serialized, cliVersion, secrets)),
+  });
+
+  return handleResponse(response);
+}
+
+function gqlBody(
+  tg: String,
```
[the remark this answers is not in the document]
afmika: I had issues with `reqwest` and `async`, but I guess we can construct the request body from there.

whiz.yaml
```diff
@@ -54,7 +54,8 @@ typegraph:
     wasm-tools component new target/wasm/wasm32-unknown-unknown/debug/typegraph_core.wasm -o $WASM_FILE
 
     rm -rf typegraph/node/sdk/src/gen typegraph/node/sdk/dist
-    jco transpile $WASM_FILE -o typegraph/node/sdk/src/gen
+    jco transpile $WASM_FILE -o typegraph/node/sdk/src/gen --map print=../host/print.js read-file=../host/read_file.js write-file=../host/write_file.js expand-glob=../host/expand_glob.js
```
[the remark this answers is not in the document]
afmika: Definitely, I actually tried that first but I had a hardtime making it work with more than one export, the jco generated code has import bugs. But I will see.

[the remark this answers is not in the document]
zifeo: try upgrading jco to 1 and explore meta v0.1.0 I believe we had it set at that point

Yohe-Am (approved):
Looking good and clean. The functionality duplicated amongst the lang specific sdks seems simple enough but if more's to be added there, we can always rely on some test suites to check conformance. 

Also, to put things into context, am I correct in my understanding the sdk code is only expected to run on dev machines and maybe CI? The most constrained env, I recall correctly, is the browser based typegraph authoring idea which would benefit if the `meta-cli` wasn't a necessity.

zifeo (approved):
lgtm! @Yohe-Am I added a bit of context [here](https://linear.app/metatypedev/issue/MET-383/sdk-move-post-processing-functions-to-the-typegate) to be discussed/completed

meta-cli/src/utils/mod.rs
```diff
@@ -61,8 +59,8 @@ pub fn plural_suffix(count: usize) -> &'static str {
 
 #[derive(Debug, Clone)]
 pub struct BasicAuth {
```
Natoandro, line 61: Why not just re-export?
```
pub use common::node::BasicAuth;
```

typegraph/node/sdk/src/tg_deploy.ts
```diff
@@ -0,0 +1,54 @@
...
+export async function tgDeploy(
+  typegraph: TypegraphOutput,
+  params: TypegraphDeployParams,
+) {
+  const { baseUrl, cliVersion, secrets, auth } = params;
+  const { serialized } = typegraph;
+
+  const headers = new Headers();
+  headers.append("Content-Type", "application/json");
+  if (auth) {
```
Natoandro, line 28: Why not defaults to the credentials provided in the `metatype.yaml`? We could pass them through environment variables...

Same for the typegate url and secrets...

Natoandro (commented):
@afmika Sorry for the late review :sweat_smile:

typegraph/node/sdk/src/tg_deploy.ts
afmika: My assumption was that `metatype.yaml` is for a better dx with the cli, we could add support for that though, but I am not sure if the env variable approach is the proper way to do it in this context, we can instead manually load the yml file from the typegraph and use its content in memory.

meta-cli/src/utils/mod.rs
afmika: We have an `From<common::node::BasicAuth>` right after instead.

PR #577, opened 2024-02-03
fix(ci): missing ts modules from `typegate` image

Yohe-Am:
Adds a check step to the `typegate`'s `Dockerfile` and converts the `.dockerignore` to be a whitelist.

#### Motivation and context

The image for 0.3.3 has some files missing.

#### Migration notes

_No changes required_

### Checklist


dev/Dockerfile
```diff
@@ -128,13 +128,27 @@ COPY --from=builder /app/target/release/typegate /bin/typegate
...
 COPY --from=builder --chown=nonroot:nonroot /app/typegate/deno.lock ./typegate/
 COPY --from=builder --chown=nonroot:nonroot /app/typegate/tmp ./
 
+# distroless-nonroot is by default the nonroot user
+# which prevents us from removing the /deno bin down below
+USER root
+
+# we temporarliy copy the deno bin and run the type checker
+# to make sure all required ts files are found in the image
+COPY --from=deno-bin --chown=nonroot:nonroot /usr/bin/deno /bin/deno
```
zifeo, line 145: This likely will have an effect on the image size, maybe running that operation in another target in the docker file would save us that?

.github/workflows/release.yml
```diff
@@ -117,10 +117,16 @@ jobs:
           cargo install cross
 
           cross build --release --locked --package meta-cli --target ${{ matrix.target }} --features typegate -v
-          tar czvf meta-cli-full-${{ steps.latest-tag.outputs.tag }}-${{ matrix.target }}.tar.gz "target/${{ matrix.target }}/release/meta${{ matrix.suffix }}"
+          # we cd next to the file so that the it's parent dirs
```
zifeo, line 120:
```suggestion
          # we cd next to the file so that the its parent dirs
```

typegate/src/runtimes/deno/worker.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright Metatype OÜ, licensed under the Elastic License 2.0.
 // SPDX-License-Identifier: Elastic-2.0
 
+/// <reference lib="deno.worker" />
```
zifeo, line 4: can you explain that addition 🙏 ?

Yohe-Am: This is a marker for `deno_lsp` to bring the `WorkerOptions` types into the global scope in this module including and extra` Worker` ctor arguments. `deno check` seemed to be throwing otherwise.

dev/Dockerfile
Yohe-Am: Oof, forgot about that. Using a separate `check` target does seem to help! Down to 347 megs from 488. Surprised it's 373 tbh. The `typegate` bin is apprently 192 megs now. (and the fat version of the `meta-cli` is 194 :/). I guess 347 makes sense when you account for the 80 megabyte `libwasmedge.so`, all the pre-downloaded `ts` files...etc.

typegate/src/runtimes/deno/worker.ts
Yohe-Am: Have added a comment explaining this to the file.

dev/Dockerfile
Yohe-Am: Wait a minute, `meta-cli` fat is only 70 megs gzipped. The 347 number's for the metatype image as a whole. Oci does gzip the tars, right?

Yohe-Am: Does it only compress the images in transit? Because the deno binary making a 150 megabyte difference only makes sense if the image layer tars are stored on disk uncompressed.

Yohe-Am: Websearch seems to confirm this. I suppose having to decompress large tarballs everytime an local image is run is not exactly efficient.

zifeo: @Yohe-Am yeah, we would need to take a different approach (we shall test the copy and paste of every binary) or try with https://github.com/slimtoolkit/slim - but not urgent, 300Mo uncompressed is fine

zifeo (approved):

PR #587, opened 2024-02-10
feat: Parameter transformation

Natoandro:
Enable parameter transformation with the `.apply()` method.
It has more or less the same logic as `.reduce()` with the ability to flatten the input type.

#### Motivation and context

This feature enables simpler APIs (input types) on top of runtimes (e.g.: prisma).

#### Migration notes

_No changes needed_.

### Checklist


typegate/src/engine/planner/parameter_transformer.ts
```diff
@@ -0,0 +1,161 @@
...
+  ): string {
+    const varName = this.#createVarName();
+    this.#collector.push(`const ${varName} = {};`);
+
+    for (const [key, node] of Object.entries(fields)) {
+      const path = this.#path;
+      path.push(key);
+      const propVarName = this.#compileNode(node);
+      path.pop();
+      this.#collector.push(`${varName}["${key}"] = ${propVarName};`);
```
afmika, line 104: Can't we do these code gynmastics without strings? This approach seems susceptible for injection as it is directly tied to the arg planner.

typegate/tests/params/apply_test.ts
```diff
@@ -0,0 +1,104 @@
...
+      })
+      .on(e);
+  });
+
+  await t.should("work with renamed params", async () => {
+    await gql`
+      query {
+        flattened(a1: 1, a2: 2, b11: 3, b12: 4, b2: 5) {
+          a { a1 a2 }
+          b { b1 { b11 b12 } b2 }
```
afmika, line 32: Neat 👍

typegate/src/engine/planner/parameter_transformer.ts
Natoandro: Thanks :+1:

typegate/tests/params/apply.py
```diff
@@ -0,0 +1,98 @@
...
+                {
+                    "a": t.list(t.integer()),
+                }
+            )
+        ).apply(
+            {
+                "a": [
+                    g.as_arg("first"),
+                    g.as_arg("second"),
+                ]
```
afmika, line 95: Looking great so far!
Maybe add tests for `g.as_arg("..")` at the end the following edgecase edgecases at the end:
1. A nested arrays at different level `[[g.as_arg("A")], g.as_arg("B")]` 
2. Nested arrays from some items of an array of objects `[ [ { a: {b: [ g.as_arg("B") ] } ], ... ]`
They both look tricky imo.

Yohe-Am (dismissed):

Yohe-Am (approved):

PR #593, opened 2024-02-16
feat(sdk): from_random injection

destifo:
This change includes changes in StringFormats(added some string formats), logic to provide random values for type nodes and tests to validate the changes. 
The changes are mostly in the typegraph sdk. 

#### Motivation and context

This feature enables the user to inject random values for a field(**Type Node**) when defining a **Typegraph**.

#### Migration notes
_No changes needed_.

### Checklist


destifo:
There is one change I would like to apply on this impl and that is for the from_random() not to accept any arguments as there is no need for it, for now I am passing "rand"(which is basically useless), but I included it cause generic injection need a parameter and that's why. I am thinking of creating a new injection(still generic) but that doesn't accept a parameter, and wanted to know what you think...

typegate/src/typegraph/mod.ts
```diff
@@ -349,6 +351,23 @@ export class TypeGraph {
     );
   }
 
+  getRandom(
+    schema: TypeNode,
+  ): number | string | null {
+    const tgTypes: TypeNode[] = this.tg.types;
+    const chance: typeof Chance = new Chance(1);
```
Natoandro, line 358: The seed should be configurable on the typegraph...
With something like:
```python
g.configureRandomInjection(seed=1)
```
It is mostly needed for the tests to be deterministic, but I think it should be random by default.

Natoandro (changes requested):
Nice :+1: 

Some suggestion :)

Natoandro:
> There is one change I would like to apply on this impl and that is for the from_random() not to accept any arguments as there is no need for it, for now I am passing "rand"(which is basically useless), but I included it cause generic injection need a parameter and that's why. I am thinking of creating a new injection(still generic) but that doesn't accept a parameter, and wanted to know what you think...

I think no arg is great!

typegate/tests/random/injection/random_injection.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@typegraph()
+def random_injection(g: Graph):
+    pub = Policy.public()
+    deno = DenoRuntime()
+
+    user = t.struct(
+        {
+            "id": t.uuid().from_random("rand"),
```
afmika, line 14: What does the `"rand"` string do?

afmika:
> > There is one change I would like to apply on this impl and that is for the from_random() not to accept any arguments as there is no need for it, for now I am passing "rand"(which is basically useless), but I included it cause generic injection need a parameter and that's why. I am thinking of creating a new injection(still generic) but that doesn't accept a parameter, and wanted to know what you think...
> 
> I think no arg is great!

I second this, I think config is enough.

typegate/src/typegraph/mod.ts
destifo: Ok, I will work on doing that.

libs/common/src/typegraph/types.rs
```diff
@@ -116,8 +117,16 @@ pub enum StringFormat {
...
     DateTime,
-    // Path,
     Phone,
+    Name,
+    Password,
+    Gender,
+    Firstname,
+    Lastname,
+    Prefix,
+    Profession,
```
Yohe-Am, line 127: thought: I've some reservations about adding more here beyond what [json schema](https://json-schema.org/understanding-json-schema/reference/string#built-in-formats) provides. It's a very natural for your implementation but anything you add to the API, someone will depend on and will have to maintained. And unlike the new ones, I believe the rest of the formats have very standard type checking semantics. Localization and more makes it tricky to add any kind of sensible functionality here.

typegate/src/engine/typecheck/common.ts
```diff
@@ -45,6 +45,16 @@ const formatValidators: Record<StringFormat, FormatValidator> = {
   phone: validator.isMobilePhone, // ??
   date: validator.isDate,
   "date-time": validator.isISO8601,
+  name: (value: string) => validator.isAlpha(value.replace(/\s+/g, "")),
+  password: (_: string) => true,
+  gender: (value: string) =>
+    ["male", "female", "other"].includes(value.toLowerCase()),
+  firstname: validator.isAlpha,
```
Yohe-Am, line 52: issue: some folks might have numbers in their names or non-latin utf-8 chars. This goes for a lot of these here.

libs/common/src/typegraph/types.rs
Yohe-Am: @afmika @Natoandro @zifeo what do y'all think?

Yohe-Am (commented):
And yeah, if the current impl stands, the arg on `from_random` doesn't seem necessary. The current pick random generators don't have any parameters it seems.

libs/common/src/typegraph/types.rs
Yohe-Am: Also, ime, random name and profession and gender generation is rarely used in production but to generate test data. Not sure if that usecase justifies adding this capability to the typegraph.

afmika: Makes sense, I believe we should keep this as simple as possible as it is mostly intended for testing. We should avoid any locale specific generators.

typegate/src/engine/typecheck/common.ts
destifo: yeah, that makes sense, I will ensure that if the StringFormat won't get removed...

libs/common/src/typegraph/types.rs
afmika: Also, things like genders can always be described with a `t.enum(...)`.

destifo: Ok, I will wait for some reviews and push an update.

Natoandro: Yeah. I agree with @Yohe-Am on this.

destifo: sure @Natoandro, I will make the changes.

Yohe-Am: Yeah, any idea what direction you're going to take it @destifo ? The current impl seems to be appropriate in a lot of ways.

Yohe-Am: A solution might be to restrict these features/semantics just to the `from_random` feature. I.e. args on `from_random`. It's still technically becomes part of the typegraph spec though.

destifo: so these features are still possible with no need to provide any arguments to the `from_random` method, as you can add them by using the `config` parameter(check the `random/random_injection/random_injection.py` test) in the typenode definition. I added these **StringFormats** to add extra flexibility to the user. so the direction I am thinking of right now is to remove the newly added StringFormats because generating random values of different types is still possible using `config` parameter of the typenode. what do you think @Yohe-Am.

typegate/src/typegraph/mod.ts
destifo: @Natoandro, I was trying to provide a gateway to configure the seed from the typegraph as per your suggestion like this
```python 
g.configureRandomInjection(seed=1)
```
The similar implementation that I found to what I am trying to achieve is `g.inherit()` which calls the `InheritDef()` class. But I wasn't able to extract as much as info I need to add the method to configure the seed for random injection as 'g.inherit()' is used in a different use case. Are there any similar implementations that have the same use case(adding configs from the typegraph) as `g.configureRandomInjection(seed=1)` that I can look into?

afmika: @destifo you need to save the seed before finalizing the typegraph so maybe you can use a store similarly this value
https://github.com/metatypedev/metatype/blob/main/typegraph/core/src/global_store.rs#L63

destifo: oww ok, so I should be saving the configured seed value inside the store as a field @afmika? just like `deploy_cwd_dir`?

zifeo: I believe you could also use g.seed(0) which might be used by different runtimes as well

destifo: Ok @zifeo, I will look into the details and apply your suggestions.

typegate/src/engine/typecheck/common.ts
destifo: Done.

typegate/src/typegraph/mod.ts
destifo: Done.

typegate/tests/random/injection/random_injection.ts
```diff
@@ -31,6 +31,9 @@ typegraph("random_injection", (g: any) => {
   const pub = Policy.public();
   const deno = new DenoRuntime();
 
+  // Configure random injection seed value or the default will be used
+  g.configureRandomInjection(1);
```
afmika, line 35: Nice!
I would suggest adding more context to the parameters, like `g.configureRandomInjection({seed: 1});` wdyt

typegate/src/runtimes/random.ts
```diff
@@ -78,48 +78,93 @@ export class RandomRuntime extends Runtime {
...
+
+function calculateCheckDigit(ean: string) {
+  const digits = ean.split("").map(Number);
+
+  let sum = 0;
+  for (let i = 0; i < digits.length; i++) {
+    sum += (i % 2 === 0) ? digits[i] : digits[i] * 3;
   }
+
+  const checkDigit = (10 - (sum % 10)) % 10;
```
afmika, line 167: shouldn't `10 - (sum % 10)` be enough? since `0 <= anything_positive % 10 <= 9`

typegate/tests/random/injection/random_injection.ts
destifo: Good point! I think it's good to add the parameters to more context and semantics to the code usage, will push an update

typegate/src/runtimes/random.ts
destifo: Done

libs/common/src/typegraph/mod.rs
```diff
@@ -33,6 +33,7 @@ pub struct Typegraph {
     pub runtimes: Vec<TGRuntime>,
     pub policies: Vec<Policy>,
     pub meta: TypeMeta,
+    pub random_seed: Option<u32>,
```
Natoandro, line 36: I suggest to move this in the `meta` field (`TypeMeta` struct).

Natoandro (dismissed):
Great!

typegate/src/runtimes/random.ts
Natoandro: Actually `10 - (sum % 10)` is between `1` and `10` inclusive...

afmika: yes indeed, my bad, that's on me.

typegraph/node/sdk/src/types.ts
```diff
@@ -325,6 +331,54 @@ export function datetime() {
...
+  return string({ format: "profession" });
+}
+
+export function city() {
+  return string({ format: "city" });
+}
+
+export function country() {
+  return string({ format: "country" });
+}
```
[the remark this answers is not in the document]
destifo: no, this is a dead code, my bad, I will go over all my changes one more time and check for dead codes and as such ... thanks for pointing it out.

typegate/tests/random/injection/random_injection.ts
destifo: Done

libs/common/src/typegraph/mod.rs
destifo: ok, I guess that makes more sense. will do that.

destifo: Done.

Natoandro (approved):

zifeo:
@destifo bravo!

PR #595, opened 2024-02-21
feat!: Nested context query

Natoandro:
- Revert context flattening
- Enable jsonpath-like key to access nested object fields or array items on the context.

#### Migration notes

If you access the context directly in your application (through the token), access to nested fields shall be updated.
E.g. the expression `context["profile.id"]` have to turned to `context.profile.id`.

### Checklist


zifeo (commented):
Nice! The migration notes should be filled right as there is something breaking (also the "!" in the title)?

Yohe-Am (approved):

Natoandro:
> @zifeo Actually it is not breaking. Only the internals have changed. User code does not need to be changed...

I take this back :sweat_smile:

zifeo (dismissed):

Yohe-Am (approved):

PR #598, opened 2024-02-23
feat(deno/sdk): native function embedding in typescript

afmika:
Add support for function or lambda definition typescript sdk for `deno.func` similarly to how `python.from_def` in python sdk works.

#### Motivation and context

Providing a string is a bit impractical and counter-intuitive espcially when the sdk language matches with runtime's language.

#### Migration notes

No changes needed.

### Checklist


zifeo (approved):

zifeo (approved):

typegraph/node/sdk/src/runtimes/deno.ts
```diff
@@ -26,7 +26,7 @@ interface PredefinedFuncMat extends Materializer {
 }
 
 export interface DenoFunc {
-  code: string;
+  code: string | Function;
```
Natoandro, line 29: The name `code` would suggest that it is a string...

typegraph/node/sdk/src/runtimes/deno.ts
```diff
@@ -38,6 +38,26 @@ export interface DenoImport {
   effect?: Effect;
 }
 
+function stringifyFn(code: string | Function) {
+  if (typeof code == "function") {
+    const source = code.toString();
+    const namedFnMatch = source.match(/function\s*(\*?\s*[a-zA-Z0-9_]+)/);
+    if (namedFnMatch) {
+      const [, name] = namedFnMatch;
+      if (name.replace(/s/g, "").startsWith("*")) {
```
Natoandro, line 47: Do you mean character `'s'`? Or perhaps `\s`?

afmika: Nice catch!

typegraph/node/sdk/src/runtimes/deno.ts
afmika: Yeah. right. I thought of renaming it but it will break old typegraphs that uses the deno feat.

Natoandro (approved):

PR #599, opened 2024-02-23
feat(cli): long running discovery

afmika:
Delegate serialize, deploy, undeploy, unpack work to SDK.

#### Motivation and context

Remove duplicate logic, thinking of cli as a convenience on top of the SDK.

#### Migration notes

When meta cli is used, Migration files are unpacked/resolved relative to the typegraph's path, not the process's `workdir`.

### Checklist


meta-cli/src/cli/serialize.rs
```diff
@@ -100,39 +107,39 @@ impl Action for Serialize {
...
+        //     if tgs.len() == 1 {
+        //         self.write(&self.to_string(&tgs[0])?).await?;
+        //     } else {
+        //         bail!("expected only one typegraph, got {}", tgs.len());
+        //     }
+        // } else {
+        //     self.write(&self.to_string(&tgs)?).await?;
+        // }
+
+        exit(0); // kill the server
```
Natoandro, line 142: You could gracefully shutdown the server, if needed:
`HttpServer::run` returns a [`Server`](https://docs.rs/actix-web/latest/actix_web/dev/struct.Server.html) instance from which you can get a [`ServerHandle`](https://docs.rs/actix-web/latest/actix_web/dev/struct.ServerHandle.html).

meta-cli/src/typegraph/loader/mod.rs
```diff
@@ -87,7 +97,10 @@ impl<'a> Loader<'a> {
 
     async fn load_command(&self, mut command: Command, path: &Path) -> LoaderResult {
         let path: Arc<Path> = path.into();
+        let path: Arc<Path> = path.clone().into();
```
Natoandro, line 100: Huh? Why the `clone`?

afmika: My bad, I did some tests then forgot removing this line.

examples/sample/example.ts
```diff
@@ -0,0 +1,24 @@
+import { Policy, t, typegraph } from "../../typegraph/node/sdk/dist/index.js";
+import { DenoRuntime } from "../../typegraph/node/sdk/dist/runtimes/deno.js";
+import { PythonRuntime } from "../../typegraph/node/sdk/dist/runtimes/python.js";
```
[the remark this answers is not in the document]
afmika: Will rm this file, I accidentally added it.

meta-cli/src/cli/codegen.rs
```diff
@@ -48,29 +44,29 @@ pub struct Deno {
...
-    async fn run(&self, args: GenArgs) -> Result<()> {
-        let dir = args.dir()?;
-        // try to find config file, else use default config as the options
-        // used for code generation have default values.
-        let config = Arc::new(
-            Config::load_or_find(args.config, &dir).unwrap_or_else(|_| Config::default_in(&dir)),
-        );
+    async fn run(&self, _args: GenArgs) -> Result<()> {
+        // TODO:
+        // how does this fit with the current impl?
```
Yohe-Am, line 49: Let's move this under the `meta gen` command? It'd be another module for `metagen`.

typegraph/node/sdk/src/typegraph.ts
```diff
@@ -212,25 +212,20 @@ export function typegraph(
...
-    };
-  }
-  return {
-    serialize,
+    },
     name,
-  };
+  } as TypegraphOutput;
+
+  if (Manager.isRunFromCLI()) {
```
Yohe-Am, line 223: thought: maybe supply a second function for the auto-deploying method or are there cases where one might use the same `typegraph` through tgDeploy and CLI?

Yohe-Am (approved):
You weren't kidding about that `Cargo.lock` diff. Good stuff, changes were way more contained than I expected.

meta-cli/src/cli/serialize.rs
afmika: Sorry for the late reply, but this can get a bit tricky as the handle needs to be available globally.
Hopefully an implementation like `ServerStore` should make it work.

examples/deploy/deploy.py
```diff
@@ -106,6 +111,7 @@ def deploy_example_python(g: Graph):
     # what to do with the migration files?
     base_dir = artifacts_config.prisma_migration.migration_dir
     # Convention, however if migration_dir is absolute then you might want to use that instead
-    full_path = path.join(base_dir, tg.name, item["runtime"])
+    # cwd + tg_name + runtime_name
+    full_path = path.join(base_dir, item["runtime"])
```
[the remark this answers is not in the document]
afmika: yes we can, I just wanted to emphasize that there are no assumption made for the migration path when using `tgDeploy`, the convention is made on the `cli`.

examples/deploy/deploy.mjs
```diff
@@ -93,20 +95,23 @@ tgDeploy(tg, {
     // dir: "."
   },
 }).then(({ typegate }) => {
+  // console.info(typegate);
   const selection = typegate?.data?.addTypegraph;
```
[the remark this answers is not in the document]
afmika: yes, but from the cli's point of view I considered that there shouldn't be modifications with the typegate's response as the goal was to relay the output of `tgDeploy` to cli.

Yohe-Am (approved):

typegraph/node/sdk/src/typegraph.ts
Yohe-Am: What I meant was to provide two versions of the `typegraph` function itself. The default one that auto-deploys and expects to be invoked from the CLI and another that returns the typegraph desc for `tgDeploy` to consume. The assumption here being that a user who's deploying through the CLI would probably not be able to use tgDeploy on the same `typegraph` fn invocation.

Upon further thought though, maybe test only users `tgDeploy` might want both functionality.

meta-cli/src/cli/serialize.rs
afmika: Update: making the Server available prior running `cmd.run` made it possible 👍

Natoandro (approved):

zifeo (approved):

meta-cli/src/com/server.rs
```diff
@@ -0,0 +1,163 @@
...
+#[get("/command")]
+async fn command() -> impl Responder {
+    match ServerStore::get_command() {
+        Some(command) => HttpResponse::Ok()
+            .status(StatusCode::OK)
+            .json(CLIResponseSuccess {
+                data: serde_json::to_value(command).unwrap(),
+            }),
+        None => HttpResponse::Ok()
+            .status(StatusCode::INTERNAL_SERVER_ERROR)
```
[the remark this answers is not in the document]
afmika: Actually, these errors are not really about the sdk making bad requests.
My reasoning was that `metatype.yaml` file is loaded on cli side (server), so it made sense that the config not being set was its responsibility.
Similarly with `command`, it's up to the cli (server) to make sure it's not optional.

PR #606, opened 2024-03-01
fix: esm customizer for typegate deno ext

### LINKED ISSUE #591: `meta typegate` doesn't work
The issue seems to be around the custom meta extension. At startup, deno tries to locate the `runtime.js` initializer at the path it'd seen it during compilation instead of bundling it within. 

I believe, in order to solve this, we'll need to start using snapshots. Or use them properly as the `deno` crate's using them by default, it's just that the meta extension isn't included in that snapshot. This will mean a few more changes in embedded `deno` repo as well.

Yohe-Am:
Enables v8 snapshots integration for `meta typegate` subcommand and the standalone typegate. Also bumps deno to 1.41.0.

#### Motivation and context

Fixes #591 

#### Migration notes

_No changes required_

### Checklist


Natoandro (dismissed):

typegate/engine/src/snapshot.rs
```diff
@@ -0,0 +1,26 @@
...
+// Copyright Metatype OÜ, licensed under the Mozilla Public License Version 2.0.
+// SPDX-License-Identifier: MPL-2.0
+use deno::deno_runtime::ops::bootstrap::SnapshotOptions;
+use mt_deno::deno;
+
+pub fn create_snapshot(snapshot_path: std::path::PathBuf) {
+    // let cli_snapshot_path = PathBuf::from(env::var_os("OUT_DIR").unwrap()).join("CLI_SNAPSHOT.bin");
+
+    let snapshot_options = SnapshotOptions {
+        // NOTE: keep this in sync
```
zifeo, line 13: ./dev/lock.ts can do it?

typegate/engine/Cargo.toml
```diff
@@ -3,6 +3,10 @@ name = "typegate_engine"
 edition.workspace = true
 version.workspace = true
 
+[features]
+default = []
+__runtime_js_sources = ["mt_deno/__runtime_js_sources"]
```
zifeo, line 8: any reason for prefixing with __ ?

typegate/engine/Cargo.toml
```diff
@@ -45,3 +49,5 @@ env_logger.workspace = true
 
 [build-dependencies]
 shadow-rs.workspace = true
+# we only need deno_runtime but this way, we avoid
```
zifeo, line 52: I am not clear on that one, can you explain?

dev/Dockerfile
```diff
@@ -18,8 +18,7 @@ COPY . .
 RUN cargo chef prepare --recipe-path recipe.json
 
 #
-# provides ARM binaries: https://github.com/LukeChannings/deno-arm64
-FROM lukechannings/deno:v${DENO_VERSION} AS deno-bin
+FROM denoland/deno:v${DENO_VERSION} AS deno-bin
```
zifeo, line 21: 🎉

typegate/engine/src/snapshot.rs
Yohe-Am: 🙌🏿 but I have removed all snapshotting code now.

typegate/engine/Cargo.toml
Yohe-Am: Oof, stale comment. I was experimenting on which crate to put the build script at first.

typegate/engine/Cargo.toml
Yohe-Am: That's what the deno codebase for a similar flag and I decided against deviating from that. `mt_deno` does pas such a flag to `embedded-deno` when it's set as well.

Most likely, they added the underscores to indicate it's private.

No matter, I've removed the snapshots code.

typegate/engine/src/ext.rs
```diff
@@ -51,14 +52,23 @@ deno_core::extension!(
...
     },
+    customizer = |ext: &mut deno_core::Extension| {
+        ext.esm_files.to_mut().push(
+            deno_core::ExtensionFileSource::new(
+                "ext:tg_metatype_ext/00_runtime.js",
+                include_str!("../00_runtime.js")
+            )
+        );
+        ext.esm_entry_point = Some("ext:tg_metatype_ext/00_runtime.js");
+    },
```
Yohe-Am, line 71: This is the workaround btw.

Yohe-Am:
Note: I also faced the issues I'd with the snapshot based `xtask` with the workaround (it wasn't related to snapshotting as it turns out though the snapshot setup did obscure it) but I did manage to fix it. Thus, I could probably go back to the snapshot paths at this point. I've decided against it though on the basis of compilation times.

The 2000 packages quote from last time turned out to be due to duplicate builds necessitated by build scripts (or some misuse on my part but I can't figure out where). 2000 packages not only takes way more time but adds way too much bloated artifacts which are even more painful at this point. It also seems to have made the cargo change detection more brittle, unnecessarily re-building packages leading to more artifacts. 

We'll probably want to enable snapshots at some point but I think it might be better to keep it simple until our dependency tree has stabilized.

zifeo (approved):

PR #618, opened 2024-03-08
fix: Optimize typegraph size

Natoandro:
- Hash all type data to compare them on the conversion phase in typegraph/core: remove duplicate types from type final typegraph (duplicate: same value for all the fields except for the "random" name/title).
- Skip unreferenced types in `.apply`

#### Motivation and context

Typegraph is too big sometimes.

### Checklist


dev/utils.ts
```diff
@@ -22,8 +22,8 @@ export async function runOrExit(
   }).spawn();
 
   // keep pipe asynchronous till the command exists
-  void p.stdout.pipeTo(Deno.stdout.writable, { preventClose: true });
-  void p.stderr.pipeTo(Deno.stderr.writable, { preventClose: true });
+  await p.stdout.pipeTo(Deno.stdout.writable, { preventClose: true });
```
zifeo, line 25: can you remove/update the comment or are we sure this is expected?

typegraph/core/src/types/type_def.rs
```diff
@@ -13,6 +14,7 @@ use crate::wit::core::{
 };
 use common::typegraph::{Injection, TypeNode};
 use enum_dispatch::enum_dispatch;
+use std::hash::Hash as _;
```
afmika, line 17: We have two places that requires some form of caching right.. (typegate introspection + SDK) I think investigating on the typegate side might also give more insights at what factor we can realistically reduce the type size before implemenenting it, I mean.. if the most effective reduction scheme is still slower than the actual growth as there are a lot of unique types we might have to think of alternative solutions 🤔.. also I am curious, have you already tried this approach on a heavy typegraph?

afmika (approved):
Nice!

PR #633, opened 2024-03-18
feat(sdk): .tgignore file support

afmika:
#### Motivation and context

Set what files/folders should be ignored when using the custom `expand_path` function in an external `.tgignore` file.
`.tgignore` will behave similarly to most .ignore files with basic glob syntax support.

#### Migration notes

`expand_glob` has been renamed to `expand_path`

### Checklist


Yohe-Am:
Hey, most ignore files rely on globs I believe. I definitely appreciate the more powerful regex but what do you think about how this might surprise users?

zifeo:
https://github.com/BurntSushi/ripgrep/tree/master/crates/ignore provide a conventional implementation and we likely already used that lib, that might indeed be a preferred approach?

afmika:
> https://github.com/BurntSushi/ripgrep/tree/master/crates/ignore provide a conventional implementation and we likely already used that lib, that might indeed be a preferred approach?

The sdk uses imported functions from the host, I believe this lib internally uses rust's fs functions to read the directory content.

zifeo:
@afmika let's give it a try with wasi for a half day 🙏

Yohe-Am (dismissed):
Hey, is the issue with the `test-full` job here unrelated to the changes?

PR #644, opened 2024-03-26
feat(gate,sdk): wasm mdk interface

afmika:
#### Motivation and context

Interface for mdk feature.

#### Migration notes

N/A

### Checklist


zifeo (commented):
I would love to see a helper like WasmRuntime().load("..", codegen=True) and each time there is a deployment, the codegen runs and fail the deployment if it changes? Then, we would need to see how to run that in dev mode.

Yohe-Am:
Superseded by #669 and #673 .

PR #645, opened 2024-03-27
feat(sdk):  introduce flag for disabling hashing artifacts + move hash to rust

afmika:
#### Motivation and context

The mdk codegen and typegraph mutually depends on each other (typegraph needs a concrete mdk.wasm for hashing, and for the mdk.wasm to be built, it needs type generation based on the typegraph)
Add a flag to enable processing a partial typegraph when using `meta gen` (partial == no artifact resolution).

#### Migration notes

`get_file_hash` has been moved to core sdk (under the name `hash_file`)

### Checklist


typegraph/core/src/typegraph.rs
```diff
@@ -245,6 +235,15 @@ pub fn finalize(
         Err(e) => return Err(e),
     };
 
+    // TODO: maybe remove this extra return value? tg.meta.ref_artifacts can be accessed on the sdk frontend
```
destifo, line 238: you mean by accessing it from the TgContext or from the serialized typegraph itself?

destifo (approved):
Great work all round!

typegraph/core/src/typegraph.rs
afmika: yes, accessing it directly from the the serialized typegraph unless there are use cases that I am missing.

destifo: I talked about this with @Natoandro and we thought rather than accessing it from the serialized typegraph, returning selectively the things you need as finalization result. We dropped the former approach not to introduce functionalities to parse the tgJson to typegraph object to the SDK.

Yohe-Am:
Hope you don't mind but I'm going to merge the `test-docker` fix into this PR and close the other one. Reasoning being this one fixes the `wasmedge_test` issue that the other one suffers from.

Yohe-Am (approved):

destifo (approved):

dev/utils.ts
```diff
@@ -13,16 +13,17 @@ export async function runOrExit(
...
-    stdout: "piped",
-    stderr: "piped",
+    // stdout: "piped",
+    // stderr: "piped",
     env: { ...Deno.env.toObject(), ...env },
   }).spawn();
 
-  await p.stdout.pipeTo(Deno.stdout.writable, { preventClose: true });
-  await p.stderr.pipeTo(Deno.stderr.writable, { preventClose: true });
+  // await p.stdout.pipeTo(Deno.stdout.writable, { preventClose: true });
```
zifeo, line 25: a leftover?

afmika: @Yohe-Am was this a fix or..?

Yohe-Am: This was the culprit for the `test-docker` issues or part of it anyways. `dev/env.ts` calls `deno cache` (another problematic API?) with thousands of paths in it's `argv`. I'm not sure what it is about pipes that makes it hang but it works well when it's talking directly to the `tty`.

zifeo: what about `(void) p.stdout ..` and then await the status at the end?

Yohe-Am: Hmm, I was actually a bit confused as to why it was piped in the first place. It only seems to be streaming it to stdout and nowhere else. In that case, shouldn't just let it inherit the std files?

zifeo: @Yohe-Am I do no recall, maybe this is not needed anymore but I recall something required that to the log when they initially switched to that new api

PR #646, opened 2024-03-27
test: use local npm registry for tests

Natoandro:
Use verdaccio local npm registry for tests:
- The `@typegraph/sdk` package is published to the local npm registry, and can now be consumed like any npm package from Nodejs or Deno.

#### Motivation and context

...

#### Migration notes

_N/A_

### Checklist

tg-sdk-verdaccio


typegraph/node/sdk/src/runtimes/index.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright Metatype OÜ, licensed under the Mozilla Public License Version 2.0.
+// SPDX-License-Identifier: MPL-2.0
+
+export { WasmEdgeRuntime } from "./wasmedge.js";
```
afmika, line 4: I would say that keeping the previous import style is better.
I am thinking that, for example, we could import potential utility functions `import { SomeRuntime, someUtility1, someUtility2, someType } "@typegraph/sdk/runtimes/some_rt.js"` without having to re-export them.

afmika (commented):
Good stuff 🔥
I have not thought about something like this!

typegraph/node/sdk/src/runtimes/index.ts
Natoandro: That was because I had some issues for import_map with the `npm:` specifier. I have find a workaround, so I am going to revert that...

Yohe-Am (dismissed):
Great thinking indeed!

Yohe-Am (approved):

afmika (approved):

PR #676, opened 2024-04-16
docs: improve `/reference/runtimes/`

afmika:
#### Motivation and context

Better documentation

#### Migration notes

N/A

### Checklist

- [ ] The change come with new or modified tests
- [ ] Hard-to-understand functions have explanatory comments
- [x] End-user documentation is updated to reflect the change

PR #684, opened 2024-04-18
docs:  add  examples for each command

afmika:
#### Motivation and context

Getting started with `meta` cli should be easy

#### Migration notes

None

### Checklist


Yohe-Am (dismissed):

examples/typegraphs/http-runtime.py
```diff
@@ -0,0 +1,44 @@
+# skip:start
```
destifo, line 1: Do we need to skip the imports? shouldn't we keep them as a reference incase the user imports from the wrong file/lib?

destifo (approved):

examples/typegraphs/http-runtime.py
zifeo: This was a not optimal solutions, ideally, we should have the import small so the core the typegraph is the focus point, yet having a fully complete code

website/docs/reference/runtimes/temporal/index.mdx
```diff
@@ -1,13 +1,82 @@
...
+An interesting use case is to dynamically describe the operations you want to expose, this enables reusing typegraphs accross different projects or even building a small framework around it.
+
+Here is a simple example of a typegraph that takes some value from an environment variable.
+
+<SDKTabs>
+    <TabItem value="python">
+```python
+from typegraph import t, typegraph, Policy, Graph
+from typegraph.providers.temporal import TemporalRuntime
+import os
```
zifeo, line 28: Can we externalize any valid source code so everything is testing and snapshotted?

zifeo (approved):

website/docs/reference/runtimes/temporal/index.mdx
afmika: Will fix that in #676

PR #686, opened 2024-04-19
docs: add `reference/programmatic-deployment/`

afmika:
#### Motivation and context

Add missing docs for `tgDeploy`, `tgRemove`

#### Migration notes

None

### Checklist

### DIFF
--- a/website/docs/reference/programmatic-deployment/index.mdx
+++ b/website/docs/reference/programmatic-deployment/index.mdx
@@ -1,0 +1,193 @@
+import SDKTabs from "@site/src/components/SDKTabs";
+import TabItem from "@theme/TabItem";
+
+# Programmatic deployment
+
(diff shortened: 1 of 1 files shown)


destifo (approved):
Nice! But I was wondering, what does happen when you deploy multiple typegraphs that can have conflicts in database scheme. For instance both specify a table with the same name but there is difference in the table schema(field types, names and as such). Do the typegraphs operate in a sandbox manner?

Yohe-Am:
@destifo the expectation is that each typegraph uses it's own separate db. Its best-practices in the micro-service world. If users want to violate this, they'll have to disable migration generation and maintain those themselves. (or just have one typegraph own the migrations). I.e. they'll have to manually maintain the types are equivalent.

Yohe-Am (approved):

zifeo (approved):

PR #687, opened 2024-04-21
refactor(gate): wasi 0.2 pyrt

Yohe-Am:
- Rewrites the PythonRuntime host using a `componentize-py` based component.
- Leaf through this [memo](https://hackmd.io/@SC-qT-WXTROceKYdNA-Lpg/ryyAXiQlC/edit) for a mental model.

Todo:
- [x] `PythonRuntime.import_` support #699 
- [x] ~~Add `pyrt.wasm` to release job~~ obviated by `build.rs`

Items for other PRs:
- Implemen `hostcall`

#### Motivation and context

MET-404. 

#### Migration notes

_No end-user changes required_

### Checklist


libs/pyrt_component/main.py
```diff
@@ -0,0 +1,58 @@
...
+)
+
+import json
+import types
+from typing import Callable, Any
+
+handlers = {}
+
+
+class MatWire(pyrt.exports.MatWire):
```
Yohe-Am, line 21: This module implements a lot of what `python-wasi-reactor` was doing.

libs/pyrt_component/main.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, handler_fn: Callable[[Any], Any]) -> None:
+        self.handler_fn = handler_fn
+
+    def handle(self, req: Req):
+        in_parsed = json.loads(req.in_json)
+        out = self.handler_fn(in_parsed)
+        return Ok(json.dumps(out))
+
+
+def op_to_handler(op: MatInfo) -> ErasedHandler:
```
Yohe-Am, line 47: Importing modules is yet pending. It's a bit tricky here but using importlib, it should be doable.

libs/pyrt_component/pyrt/__init__.py
```diff
@@ -0,0 +1,12 @@
+from typing import TypeVar, Generic, Union, Optional, Protocol, Tuple, List, Any, Self
```
Yohe-Am, line 1: Everything in this dir is generated from the wit file. Will remove this dir and generated when needed at the end of PR.

libs/pyrt_component/wit/pyrt.wit
```diff
@@ -0,0 +1,56 @@
+package metatype:pyrt;
```
Yohe-Am, line 1: Note, nothing in this wit if python specific. In another PR, will refactor this and the generation into `metagen`.

typegate/engine/src/runtimes/wit_wire.rs
```diff
@@ -0,0 +1,316 @@
...
+            comp.clone()
+        } else {
+            let engine = self.engine.clone();
+            let path_clone = wasm_absolute_path.clone();
+            let comp = if wasm_absolute_path
+                .extension()
+                .map(|ext| ext == "cwasm")
+                .unwrap_or_default()
+            {
+                // TODO: self manage precompilation cache
```
Yohe-Am, line 61: The componentize-py artifact doesn't do the fancy [`wasi-virt` tricks](https://github.com/bytecodealliance/componentize-py/issues/23) or dynamically loading [libpyt](https://github.com/bytecodealliance/componentize-py/issues/28) to minimize the module size. This leads to a lot of work when we compile the component for the first time. Until `componentize-py` sports those features, we can improve startup by using precompilation.

To dedup work, we can have the typegate mantain a local pre-compile cache in `tmp` which should mean that users will only ever need to pre-compile the pyrt module once during dev. 

We can also distribute the `pyrt` pre-compiled for each platform. Or alternatively, use `build.rs` to embed the `pyrt` component into the typegate after pre-compiling and compressing it.

Any thoughts on this?

dev/Dockerfile
```diff
@@ -39,8 +39,6 @@ RUN set -eux; \
     automake \
     # wasm-opt deps \
     clang \
-    ## wasmedge-sys \
-    libclang-dev \
```
zifeo, line 43: 😍

libs/pyrt_wit_wire/pyrt/imports/__init__.py
zifeo, line 1: empty init are still required here?

ruff.toml
```diff
@@ -1 +1,4 @@
 ignore = ["E501"]
+exclude = [
+  "libs/pyrt_wit_wire/pyrt"
```
zifeo, line 3: temp?

typegate/engine/src/runtimes/wit_wire.rs
zifeo: I would go for the later, which I believe makes anyway sense if we keep the wasm rt optional so they can be pulled from a oci / warg repository when needed? Is that much work?

Yohe-Am: The latter as in embed the pre-compiled runtime?

zifeo: yep, at the end of the day, people could also use the runtime as component elsewhere

typegate/engine/Cargo.toml
```diff
@@ -34,19 +34,21 @@ mt_deno.workspace = true
...
-# remove wasmedge once python runtime has been migrated
-wasmedge-sdk.workspace = true
-wasmedge-sys.workspace = true
-wasmedge-types.workspace = true
-
-wasmtime = { version = "19.0.1",  features = ["component-model"] }
+wasmtime = { workspace = true,  features = ["component-model"] }
+wasmtime-wasi.workspace = true
+# remove when bumping wasmtime to 20
+cap-std = "3"
```
afmika, line 40: After some investigations, there shouldn't probably be major changes when bumping to 20 on the type conversion stuff in case some of this depends on that. They pretty much changed how to instantiate values in a more 'non-confusing' way.

typegate/engine/build.rs
```diff
@@ -1,9 +1,67 @@
...
 
-// Copyright Metatype OÜ, licensed under the Mozilla Public License Version 2.0.
-// SPDX-License-Identifier: MPL-2.0
+use std::path::PathBuf;
 
-fn main() -> shadow_rs::SdResult<()> {
-    shadow_rs::new()
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    shadow_rs::new()?;
+
```
zifeo, line 8: I believe we cover that in a standup but worth adding the reason why we use a build.rs instead of ghjk tasks?

zifeo (commented):
Nice!

typegate/engine/build.rs
Yohe-Am: I was just doing a ghjk task at first but `build.rs` seemed more idiomatic since the pyrt.wasm file is always expected. Unlike `examples/templates`, it's a derived artifact.

So far, the compile/compression compile time bump hasn't been bad. Most of the typegate is in typescript so we rarely re-compile `typegate_engine`. We'll probably need to come up with a better scheme if we add more than one inline component.

Yohe-Am: Also, in some scenarios, we pre-compile the module for other targets.

zifeo: In the short term, we will definitely only support Python but as soon as the wasm components are well supported in ruby and php, we likely going to add support for them.

.github/workflows/tests.yml
```diff
@@ -188,8 +188,10 @@ jobs:
             runner: ubuntu-latest
           # FIXME: try macos-14 runner once all actions support it
           # docker buildx action broken as of 2024-02-09
+          - platform: linux/arm64
```
zifeo, line 191: linux/arm64 on custom Mac?

zifeo (approved):

.github/workflows/tests.yml
Yohe-Am: Yeah, that's the only we we can access that triplet. Macos can run linux containers somehow (I'm guessing simulation). 

The reason that specific job is broken right now is due to this ghjk bug for that platform. I'll try to a similar approach over there and add a linux/arm64 job on macos-14.

destifo (approved):

PR #673, opened 2024-04-12
feat(metagen): `mdk_rs` finalization

Yohe-Am:
- Finalizes the host side of the `wit_wire` implementation started in #687 .
- Fixes a few issues with the `mdk_rust` generator.

#### Motivation and context

The code generator was already in place but we the typegate had no support for the `wit_wire` interface used by the rust mdk. This PR adds that.

The two week delay is mainly due to the base work required in #669 and related PRs.

#### Migration notes

_No breaking changes on user._

### Checklist

- [x] The change come with new or modified tests
- [x] Hard-to-understand functions have explanatory comments
- [ ] End-user documentation is updated to reflect the change

PR #699, opened 2024-04-29
feat(pyrt)!: `import_function` materializer support

Yohe-Am:
Implements python runtime support for `import_function` materializers. This is an stacked PR on #673 which itself is appended on #687 for tackling MET-404.

#### Migration notes

This PR introduced a breaking change in the importer logic for python modules brought in through `pyrt.import_` and it's deps. After this PR, any path imports must be prefixed with `.`, `..`, `...` and so to point to relative path locations. This behavior is at odds with standard python which looks for relative modules and packages even if there are no dot prefixes.

### DIFF
--- a/libs/pyrt_wit_wire/main.py
+++ b/libs/pyrt_wit_wire/main.py
@@ -7,6 +7,7 @@ from wit_wire.exports.mat_wire import (
     InitArgs,
     InitResponse,
     InitError_UnexpectedMat,
+    InitError_Other,
     MatInfo,
     HandleReq,
     HandleErr_NoHandler,
@@ -17,7 +18,14 @@ from wit_wire.exports.mat_wire import (
 
 import json
 import types
-from typing import Callable, Any
+from typing import Callable, Any, Dict
+import importlib
+import importlib.util
+import importlib.abc
+import importlib.machinery
+import os
+import sys
+import traceback
 
 # the `MatWire` class is instantiated for each
 # external call. We have to put any persisted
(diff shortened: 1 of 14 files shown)


libs/pyrt_wit_wire/main.py
```diff
@@ -60,8 +77,103 @@ def op_to_handler(op: MatInfo) -> ErasedHandler:
...
+                    self._pkgs.add(dirname)
+                dirname = os.path.dirname(dirname)
+        self._mod_names = {ThePathFinder.path_to_module(path): path for path in modules}
+        self._pkg_names = {
+            ThePathFinder.path_to_module(path): path for path in self._pkgs
+        }
+
+    # Look for a spec under a certain module name
+    # https://peps.python.org/pep-0302/
+    # https://peps.python.org/pep-0451/
```
Yohe-Am, line 123: What I'm seeing in these PEPs and the docs imply the limitation/breaking change I've outlined in the PR desc is not necessary but I was not able to get it working despite hours.

Some context links:
- https://stackoverflow.com/questions/41941079/what-is-the-difference-between-sys-meta-path-and-sys-path-hooks-importer-obj
- https://stackoverflow.com/questions/41990169/how-to-use-sys-path-hooks-for-customized-loading-of-modules

typegate/tests/runtimes/python/py/hello.py
```diff
@@ -1,9 +1,10 @@
-from nested.dep import hello
+from .nested.dep import hello
```
Yohe-Am, line 1: Here's the breaking change.

libs/pyrt_wit_wire/main.py
```diff
@@ -28,12 +36,20 @@
 class MatWire(wit_wire.exports.MatWire):
     def init(self, args: InitArgs):
         for op in args.expected_ops:
-            handlers[op.op_name] = op_to_handler(op)
+            try:
+                handlers[op.op_name] = op_to_handler(op)
+            except Exception as err:
+                print(f"errororororororororr {err}")
```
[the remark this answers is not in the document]
Yohe-Am: Fixed.

PR #700, opened 2024-04-30
refactor(libs/xtask,gate): remove xtask/codegen

afmika:
Faster build time

#### Migration notes

Make sure to sync `typegate/src/types.ts` when an update is made on the typegraph schema.


Yohe-Am (approved):

destifo (approved):

Cargo.toml
```diff
@@ -37,6 +37,7 @@ reqwest = "0.11.20"
 wasmedge-sdk = {  version = "=0.13.2", features = ["standalone"] }
 wasmedge-sys = "=0.17.5"
 wasmedge-types = "=0.4.4"
+
 # https://github.com/denoland/deno_ast/blob/main/Cargo.toml
 swc_atoms = "=0.6.5"
```
zifeo, line 42: do we need to keep them?

afmika: The typescript codegen still depend on swc_*, but it should be removable once the typescript mdk finalization lands.

PR #707, opened 2024-05-03
feat(mdk): mdk python

afmika:
Mdk for python runtime

#### Migration notes

None

## Summary by CodeRabbit

- **New Features**
	- Introduced new functionalities for generating Python code based on configurations, including handling of templates and required objects.
	- Added Python script templates for defining typed functions and structured objects with comprehensive data type handling.
	- Enhanced type management and priority handling in utility functions.

- **Documentation**
	- Provided detailed summaries and documentation for new functionalities and templates.

- **Refactor**
	- Implemented new structures and methods for efficient code generation and type handling.

- **Tests**
	- Added tests for defining typegraph structures and policies in Python.

- **Chores**
	- Updated URLs in the `.ghjk/deno.lock` file to reflect new changes in the codebase.


afmika:
I ended up using class approach
 
Example output
scripts/example.py (entry point)
```python
from example_types import Struct, IdeaInput, IdeaOutput

def handle(inp: IdeaInput) -> IdeaOutput:
    # Write your logic here
    raise Exception("not implemented")

# Do not edit this function
def identity(raw_inp):
    inp: IdeaInput = Struct.instantiate(IdeaInput, raw_inp)
    return handle(inp).repr()
```

scripts/example_types.py
```python
import json
from typing import List, Dict, Optional
from dataclasses import dataclass, asdict, fields

class Struct:
    def instantiate(dt_class, val: any):
        try:
            ftypes = {f.name: f.type for f in fields(dt_class)}
            attrs = {}
            for f in val:
                fval = val[f]
                if type(fval) is list:
                    attrs[f] = [Struct.instantiate(ftypes[f].__args__[0], v) for v in fval]
                else:
                    attrs[f] = Struct.instantiate(ftypes[f], val[f])
            return dt_class(**attrs)
        except:
            return val
    def repr(self):
        return asdict(self)


@dataclass
class References(Struct):
    name: str
    code: int
    usage: float

@dataclass
class IdeaInput(Struct):
    name: str
    age: int
    authorEmail: Optional[str]
    linkedTo: List[int]
    references: List[References]

@dataclass
class IdeaOutput(Struct):
    name: str
    age: int
    authorEmail: Optional[str]
    linkedTo: List[int]
    references: List[References]
```
typegraph.py
```
from typegraph import typegraph, Policy, t, Graph
from typegraph.runtimes.python import PythonRuntime

@typegraph()
def example(g: Graph):
    references = t.struct({
        "name": t.string(),
        "code": t.integer(),
        "usage": t.float()
    }, name="References")
    idea = t.struct(
        {
            "name": t.string(),
            "age": t.integer(),
            "authorEmail": t.email().optional(),
            "linkedTo": t.list(t.integer()),
            "references": t.list(references)
        },
        name="Idea"
    )
    python = PythonRuntime()
    pub = Policy.public()
    g.expose(
        pub,
        identity=python.import_(
            idea.rename("IdeaInput"), idea.rename("IdeaOutput"),
            name="identity",
            module="scripts/example.py"
        )
    )
```

libs/metagen/src/mdk_python/static/main.py.jinja
```diff
@@ -0,0 +1,14 @@
...
+{% else %}
+from {{ type_mod_name }}_types import Struct, {{ t_input }}
+{% endif %}
+
+def handle(inp: {{ t_input }}) -> {{ t_output }}:
+    # Write your logic here
+    raise Exception("not implemented")
+
+# Do not edit this function
+def {{ fn_name }}(raw_inp):
```
zifeo, line 12: can this be moved to the other file to have the following "convention"?
- x_types = do not edit/regenerated often
- x = do what ever you want, and you import what you need from types

libs/metagen/src/mdk_python/static/types.py.jinja
```diff
@@ -0,0 +1,25 @@
+import json
+from typing import List, Dict, Optional
+from dataclasses import dataclass, asdict, fields
+
+class Struct:
```
zifeo, line 5: Does the class provide any advantage or remove limitations?

zifeo (commented):
Nice!

libs/metagen/Cargo.toml
```diff
@@ -17,6 +17,7 @@ heck = "0.5.0-rc.1"
 regex.workspace = true
 once_cell.workspace = true
 # indoc.workspace = true
+tera = { version = "1", default-features = false }
```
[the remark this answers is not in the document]
afmika: I would suggest we use this approach accross current/future implementations. This offers a consistent way of isolating/reuse templates accross the generators.
@Yohe-Am wdyt?

libs/metagen/src/mdk_python/static/types.py.jinja
afmika: I tried `.pyi` first but it does not allow a good representation of union/either types for example nor enable hinting python dict when editing (turns out [box ](https://pypi.org/project/python-box/) is not part of the std python).
We can instead put structs and other complex types into classes, the downside is the manual serialization/deserialization work but since this is not designed to be manually edited, we can use codegen as much as we want for the tricky parts.

libs/metagen/src/mdk_python/static/types.py.jinja
```diff
@@ -0,0 +1,25 @@
...
+    def instantiate(dt_class, val: any):
```
Yohe-Am, line 6: Will all subclasses need to be instantiated this way?

libs/metagen/src/mdk_python/static/types.py.jinja
zifeo: Yes, `.ipy` is quite limiting. Dataclasses should be better than boxed struct indeed 👍 but imo != classes, that might be my confusion. Is there anything that requires to have this Struct as a class and instead uses data classes all the time?

zifeo:
@CodiumAI-Agent /review

libs/metagen/src/mdk_python/static/types.py.jinja
afmika: Not really, it does not have any fields so a simple class will do the job but I guess we can do it that way for consistency.

libs/metagen/src/mdk_python/static/main.py.jinja
```diff
@@ -0,0 +1,14 @@
...
+from {{ type_mod_name }}_types import Struct, {{ t_input }}, {{ t_output }}
+{% else %}
+from {{ type_mod_name }}_types import Struct, {{ t_input }}
+{% endif %}
+
+def handle(inp: {{ t_input }}) -> {{ t_output }}:
+    # Write your logic here
+    raise Exception("not implemented")
+
+# Do not edit this function
```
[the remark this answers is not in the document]
afmika: Yeah, that's an interesting problem. The old codegen approach I believe was 'not do anything if the file already exist'. So I suppose we can instead enable force overwrite with a `--force` flag?

libs/metagen/src/mdk_python/static/types.py.jinja
afmika: This will create a `dataclass` out of a raw json object so if the underlying `dataclass` has another `dataclass` field then it should recursively deserialize it into that (through `Struct.instantiate(..)`).

libs/metagen/Cargo.toml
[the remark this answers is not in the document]
afmika: I think we can use both approach as long as we keep it clean and consistent.

libs/metagen/src/mdk_python/static/main.py.jinja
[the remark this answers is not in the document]
afmika: Actually nvm, we can put all of these in a separate file and let the user import what they need or what the have to implement.

libs/metagen/src/mdk_python/types.rs
```diff
@@ -0,0 +1,121 @@
...
+    };
+
+    if let TypeNode::Union { data, .. } = tpe {
+        visit_variants(&data.any_of)
+    } else if let TypeNode::Either { data, .. } = tpe {
+        visit_variants(&data.one_of)
+    } else {
+        panic!("union/either node was expected, got {:?}", tpe.type_name())
+    }
+}
```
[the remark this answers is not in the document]
afmika: neat review!

libs/metagen/src/mdk_python/utils.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    pub fn types_in_order(&self) -> Vec<TypeGenerated> {
+        let mut values = self
+            .map
+            .values()
+            .filter(|c| c.type_generated.is_some())
+            .collect::<Vec<_>>();
+
+        // if typing dataclass A depends on dataclass B, then B must be generated first
+        values.sort_by(|a, b| b.priority.cmp(&a.priority));
```
Yohe-Am, line 82: Neat.

Yohe-Am (approved):
Would love a demo for this but looking good. Curious how you'll tackle re-generation for the user modified files.

libs/metagen/src/mdk_python/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+        tera.add_raw_template("struct_template", include_str!("static/struct.py.jinja"))?;
+
+        let stubbed_funs = filter_stubbed_funcs(tg, &["python".to_string()])?;
+        for fun in &stubbed_funs {
+            if fun.mat.data.get("mod").is_none() {
+                continue;
+            }
+            let (_, script_path) = get_module_infos(fun, tg)?;
+            let base_path = self.config.base.path.clone();
+            let entry_point_path = if base_path.as_os_str().to_string_lossy().trim() == "" {
```
Yohe-Am, line 97: Sort of related with the above comment but yeah, type modules per module it seems. Will this mean that the Struct impl will be duplicated for each module? I see that the impl allows multiple handlers in a single module but still, maybe it might be better to be more opinionated. 

Something like, only generate a single module with all the types and the decorators. The final layout will be then:
```
gen_target_dir/
  |- add.py
  |- identity.py
  |- mdk.py
```

`add.py` and `identity.py` will then import from `.mdk`. What do you think?

Note: in the future, we will have more code than the Struct impl in there like the hostcall wrapper so this duplication will multiply

Yohe-Am (commented):
Wonderful. Some comments but not blocking and feel free to merge.

Yohe-Am:
Oh man. Am I glad to see the `test-full` timeout happening on this PR as well. I was loosing hair over here thinking it was the deadlock bug.

Yohe-Am:
Looks like something wrong with serialization of python modules looking at the logs of runs on both PRs.

libs/metagen/src/mdk_python/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+        };
+        let mut mergeable_output: IndexMap<PathBuf, Vec<RequiredObjects>> = IndexMap::new();
+
+        let mut tera = tera::Tera::default();
+        tera.add_raw_template("main_template", include_str!("static/main.py.jinja"))?;
+        tera.add_raw_template("types_template", include_str!("static/types.py.jinja"))?;
+        tera.add_raw_template("struct_template", include_str!("static/struct.py.jinja"))?;
+
+        let stubbed_funs = filter_stubbed_funcs(tg, &["python".to_string()])?;
+        for fun in &stubbed_funs {
```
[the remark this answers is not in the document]
Yohe-Am: Not sure how this would work in python but what I meant bout the graphql queries is the variables and the data returned from queries. It'd be nice to be able to convert the JSONs into the concrete types to ease interacting with them. I don't mean auto conversion through the `gql` abstraction but manually by the user which I still think is very valuable.  This'll come in very handy when things change during regeneration, breaking the static analysis.

libs/metagen/src/mdk_python/mod.rs
Yohe-Am: I see your reasoning. The fact that you common generated items will be de-duped alleviates sort of my worries. I still generally prefer one generated file over multiple (or one directory that I can just put in a `.gitignore` or `rm` with abandon). I also don't think length of generated files is of great concern but yeah, if it's all de-duped, it looks good to me.

afmika: > or one directory that I can just put in a .gitignore or rm with abandon
Actually, you get this simply by refering the same file for all materializers.

zifeo (dismissed):
🚀

destifo (dismissed):

destifo (approved):

PR #733, opened 2024-05-23
feat: Move all the configs to one single file

Natoandro:
- Move all the configs to one single file
- Remove some specific configs from the global config variable and make them accessible on the `Typegate` instance, to improve test configurability.

#### Migration notes

- `SYNC_REDIS_PASSWORD` has been removed, can only be set on the `SYNC_REDIS_URL`.


typegate/src/sync/config.ts
```diff
@@ -1,135 +1,135 @@
...
+// export async function syncConfigFromEnv(
+//   sources: ConfigSource[],
+// ): Promise<SyncConfig | null> {
+//   const rawObjects = sources.map((source) => {
+//     switch (source) {
+//       case "vars":
+//         return mapKeys(
+//           Deno.env.toObject(),
+//           (k: string) => k.toLowerCase(),
+//         );
```
zifeo, line 127: comments on purpose?

Natoandro: Will be removed...

whiz.yaml
```diff
@@ -19,7 +19,7 @@ gate1: &tp
     REDIS_URL: "redis://:password@localhost:6379/0"
     TG_SECRET: "a4lNi0PbEItlFZbus1oeH/+wyIxi9uH6TpL8AIqIaMBNvp7SESmuUBbfUwC0prxhGhZqHw8vMDYZAGMhSZ4fLw=="
     TG_ADMIN_PASSWORD: "password"
-    LD_LIBRARY_PATH: /data/home/metatype/.ghjk/envs/default/shims/lib
```
zifeo, line 22: is that wanted?

Natoandro: The path absolute. It was specific to his setup when he added it there...

zifeo: Yeah, should it be `./.ghjk/envs/default/shims/lib` instead?

Yohe-Am: Yeah, this is no longer necessary since #754 .

Yohe-Am (approved):
Good diff.

Yohe-Am (approved):

PR #736, opened 2024-05-27
chore(docs): replace term materializer with function for user facing concepts

afmika:
Materializer and function might introduce confusion as they are pretty much the same thing from the user point of view, one can be defined in terms of the other.

#### Migration notes

None

## Summary by CodeRabbit

- **Documentation**
  - Updated terminology from "materializers" to "functions" across various documentation files to reflect a semantic shift and provide clearer descriptions.
  - Improved clarity in descriptions of custom functions, runtimes, and their roles in the Metatype computing model.
  - Corrected typos and refined explanations in multiple guides and reference documents.

These changes enhance the readability and consistency of our documentation, making it easier for users to understand and implement the features and concepts within the system.


afmika:
@coderabbitai review

zifeo (approved):
I suppose we may miss some rephrasing but that's good enough for now

afmika:
> I suppose we may miss some rephrasing but that's good enough for now

I tried to fix/remove a few things but I guess that's expected since most had the materializer concept in mind, functions was rather used to the node/type.

destifo (approved):
I wonder though what this means to the abbreviation "MDK"

afmika:
> I wonder though what this means to the abbreviation "MDK"

We should keep that, I think we can define materializer in the MDK section instead as it would actually mean something concrete to the user there.

PR #737, opened 2024-05-27
chore: bump deno to 1.43.6

Yohe-Am:
Update deno to 1.43.6 and make requisite changes.

Required because of dep conflicts with latest lade-sdk.

#### Migration notes


dev/test.ts
```diff
@@ -105,7 +106,7 @@ const env: Record<string, string> = {
   "TMP_DIR": tmpDir,
   "TIMER_MAX_TIMEOUT_MS": "30000",
   "NPM_CONFIG_REGISTRY": "http://localhost:4873",
-  "PATH": `${Deno.env.get("PATH")}:${join(projectDir, "target/debug")}`,
+  "PATH": `${join(projectDir, "target/debug")}:${Deno.env.get("PATH")}`,
```
Yohe-Am, line 109: This line took me 3 hours lol.

typegate/tests/typecheck/__snapshots__/typecheck_test.ts.snap
```diff
@@ -112,43 +99,6 @@ function validate_4(value, path, errors, context) {
     }
   }
 }
-return validate_66_1;
-"
-`;
-
-snapshot[`typecheck 2`] = `
```
Yohe-Am, line 119: All of these are gone.

Yohe-Am (commented):
Hey, so this PR exhibits a bug I can't seem to get the bottom of. The snapshot tests for `typecheck_test.ts` appear to be broken. You'll notice most of snapshots previously in the file were discarded. This wasn't intentional is precisely the issue. Some observations:

- The subset of snapshots of that are kept is different on each `--update` run.
- This means subsequent runs of the test will break because snapshots are missing.
- The actual validation logic seems to be sound as far as I were able to printf all the value/error pairs.
- There does seem to be a change introduced in this PR that required the snapshot updates. Functional or not, the generated code is different and it shouldn't be. Not sure what parts of v8/Deno are responsible. 
- I've tried updating to the latest version of the deno std.
- It appears this is the only snapshot test suffering from this bug.

I spent most of today trying to fix this. Not sure where to go from here but I'll try to do more experiments tmw.

Cargo.toml
```diff
@@ -36,14 +37,14 @@ clap = "=4.4.17"
...
-swc_ecma_transforms_typescript = "=0.186.17"
-swc_ecma_visit = "=0.96.17"
+# update from https://github.com/denoland/deno_ast/blob/main/Cargo.toml
+swc_atoms = "=0.6.7"
+swc_common = "=0.33.26"
+swc_ecma_ast = "=0.113.4"
+swc_ecma_codegen = "=0.149.1"
+swc_ecma_parser = "=0.144.1"
+swc_ecma_transforms_typescript = "=0.189.1"
+swc_ecma_visit = "=0.99.1"
```
afmika, line 47: These are only related to the typescript codegen right?
I think we talked about removing these along with the current typescript codegen at some point as these are going to be replaced by the full mdk rewrite anyways.

dev/test.ts
zifeo: jz... might add a comment for the next change there?

zifeo (commented):
lgtm

Yohe-Am:
Huh, the `python_sync_test` no longer takes 10 minutes to run but something else is blowing up the test suite runtime. I'll disable the rust cache for now but this ought to be investigated next week.

typegate/tests/utils/test.ts
```diff
@@ -324,22 +321,23 @@ export class MetaTest {
     return true;
   }
 
-  async assertSnapshot(...params: AssertSnapshotParams): Promise<void> {
```
Natoandro, line 327: wth, this function has always been `async` and never `await`ed! :thinking:

Natoandro (dismissed):

Cargo.toml
Yohe-Am: Handled in #739 .

zifeo (approved):

PR #739, opened 2024-05-30
feat(mdk)!: `mdk_typescript`

Yohe-Am:
- Implements the `mdk_typescript` code generator for typescript type inference on Deno runtime external modules.
- Ports the very simple generator already present in meta-cli.
- Removes old codegen from cli and sdk.

#### Migration notes

- Metagen section of `metatype.yaml` has changed. Targets are now lists instead of maps, items no sporting `generator` field instead of key acting as ref to generator.
- (sdk) WasmRuntime's `fromExport` method has been renamed to `export` to make it more uniform to handler.
- (sdk) WasmRuntime `export` and `handler` method's now expect handler/func name under `name` instead of `func`. 
- (sdk) `codegen` flag has been removed from `ArtifactsConfig` object.
- (cli) `gen mod/mdk` has been simplified to just `gen` as the previous mod option is no longer avail. 

---


typegate/tests/metagen/metagen_test.ts
```diff
@@ -160,3 +169,258 @@ Meta.test("Metagen within sdk", async (t) => {
...
+                to2: { // cycles2
+                  // cycles2/variant cycles1
+                  to2: null,
+                  phantom1: "phantom",
+                },
+              }],
+            },
+          },
+        } as Record<string, JSONValue>,
+      },
```
Yohe-Am, line 284: @Natoandro this is the query the breaks the typegate...

typegate/tests/metagen/typegraphs/identities.py
```diff
@@ -0,0 +1,198 @@
...
+                    t.list(t.enum(["grey", "beige"])).rename("branch4"),
+                    t.integer(),
+                    t.string(),
+                    t.email().rename("branch4again"),
+                ]
+            ),
+            "list": t.list(t.string()),
+        }
+    ).rename("composites")
+    cycles1 = t.struct(
```
Yohe-Am, line 50: These are the cyclical types that break the typegate.

Yohe-Am (commented):
Sorry for another large PR ya'll. Didn't know how I could break this down.

zifeo (commented):
Looks good, should the swc dependencies also be removed and also the post processing inside the typegate?

Yohe-Am:
Another usecase was unfortunately discovered for the typescript crates [here](https://github.com/metatypedev/metatype/blob/main/typegate/engine/src/runtimes/deno_rt.rs). To compile user provided typescript modules to javascript. (Deno also makes use of `swc`).

Note, these crates are no longer built by the thin cli.

zifeo:
Yeah, that what I meant by the post processing on the type gate side, I believe there is no usage anymore: https://github.com/metatypedev/metatype/blob/8dde1faca85547a025c1bd2e580b434ec229636b/typegate/src/postprocess.ts#L4

libs/metagen/src/mdk_python/types.rs
```diff
@@ -45,7 +44,9 @@ pub fn visit_type(
         TypeNode::Union { .. } | TypeNode::Either { .. } => {
             visit_union_or_either(tera, memo, tpe, tg)?.hint
         }
-        _ => bail!("Unsupported type {:?}", tpe.type_name()),
+        // TODO: base64
+        TypeNode::File { .. } => "str".to_string(),
+        TypeNode::Any { .. } => "Any".to_string(),
```
afmika, line 49: Maybe we can use a simple alias? `Base64String = str`

afmika (approved):
Good stuff 👍

typegate/tests/metagen/typegraphs/identities.py
afmika: Does it break at validation or.. ?

Yohe-Am: It 500s the typegate.  Nested union/either should have been flattened.

libs/metagen/src/mdk_python/types.rs
Yohe-Am: Yeah, this would need specialized handling. Base64 a 25MiB file wouldn't make sense. Maybe exposing the file as a preopen to the modules? Not sure.

meta-cli/src/typegraph/loader/discovery.rs
```diff
@@ -85,6 +85,10 @@ impl TryFrom<&Path> for ModuleType {
     fn try_from(path: &Path) -> std::result::Result<Self, Self::Error> {
         match path.extension() {
             Some(ext) if ext == "ts" => Ok(ModuleType::Deno),
+            Some(ext) if ext == "mts" => Ok(ModuleType::Deno),
+            Some(ext) if ext == "js" => Ok(ModuleType::Deno),
+            Some(ext) if ext == "mjs" => Ok(ModuleType::Deno),
+            Some(ext) if ext == "cjs" => Ok(ModuleType::Deno),
```
Natoandro, line 91: We should definitely rename this enum variant...

Natoandro (approved):

PR #746, opened 2024-06-01
chore: test buildjet

zifeo:
- improve CI speed and cache size


destifo (dismissed):
🔥 curious how much speed gain we will get

Natoandro (dismissed):

Yohe-Am:
For the `lint-compat` jobs, they don't have the new `sccache` step it seems. Probably because they didn't originally have a `rust-cache` step which I decided to exclude out of cache exhaustion concerns. I've added the fix.

PR #755, opened 2024-06-15
fix: missing typegraphs

zifeo:
- fix the typegraphs that were incorrectly formatted

### DIFF
--- a/examples/typegraphs/microservice-orchestration.py
+++ b/examples/typegraphs/microservice-orchestration.py
@@ -1,7 +1,7 @@
 # skip:start
 from os import environ
 
-from typegraph import typegraph, Policy, t, Graph
+from typegraph import Graph, Policy, t, typegraph
 from typegraph.graph.params import Cors
 from typegraph.runtimes import DenoRuntime, GraphQLRuntime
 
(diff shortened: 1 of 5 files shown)


examples/typegraphs/microservice-orchestration.py
```diff
@@ -25,17 +25,15 @@ def team_a(g: Graph):
...
-#     g.expose(
-#         public,
-#         version=deno.static(t.integer(), 12),
-#         record=deno.static(t.struct({"weight": t.integer()}), {"weight": 100}),
-#     )
+@typegraph(
+    # skip:next-line
+    cors=Cors(allow_origin=["https://metatype.dev", "http://localhost:3000"]),
+)
+def team_b(g: Graph):
```
destifo, line 32: The error on the test failure is different, but shouldn't the ts and python version of typegraph have the same name?

destifo (changes requested):

destifo:
So the missing typegraphs issue came from `stderr` of the ruff formatting command not being handled?

examples/typegraphs/microservice-orchestration.py
zifeo: I see them having the same name, can you point at where they differ?

zifeo:
@destifo quite unsure yet, but at least we would have the debug logs for that now - more knowledge, better idea where the failure is :)

examples/typegraphs/microservice-orchestration.py
destifo: overlooked the second typegraph on my part, all looks good.

destifo (approved):
lgtm

PR #760, opened 2024-06-19
refactor(sdk)!: back to deno + jsr exploration

afmika:
Try reverting back to deno runtime for the typescript sdk in hope of making the dx easier.
Hosting a custom node/npm project adds more layer of indirection which may result in cryptic issues sometimes.

This should also facilitate publishing on jsr although additional work are still required on the `jco` codegen side.

#### Migration notes

N/A


Yohe-Am (dismissed):

meta-cli/src/deploy/actors/task/command/typescript.rs
```diff
@@ -17,12 +17,17 @@ pub(super) async fn get_raw_command(path: impl AsRef<Path>) -> Result<Command> {
         TsRuntime::Deno => {
             log::debug!("loading typegraph using deno");
             let mut command = Command::new("deno");
+
+            command.arg("run");
+            if let Ok(import_map) = std::env::var("MCLI_DENO_IMPORT_MAP") {
```
Yohe-Am, line 22: **thought**: instead of a specialized param like this, wouldn't it be better to have a `MCLI_LOADER_TS`? If they want to specify an import map, they can set it to `deno --import-map=... --other-flag=...`.

Yohe-Am (dismissed):

meta-cli/src/deploy/actors/task/command/typescript.rs
afmika: yeah that's a good idea. A question though, wouldn't that look like `MCLI_LOADER_TS="deno run -A run"` for example? Which sort of implies that we need to document somewhere it is the command prefix so the filename comes after.. But I guess we can introduce templates with predefined placeholders such as `{filename}` or `{cwd}`.

.github/workflows/release.yml
```diff
@@ -186,16 +186,22 @@ jobs:
...
           [ ${{ github.event_name == 'workflow_dispatch' && inputs.ovewriteArtifacts }} == 'true' ] \
             && pnpm publish --no-git-checks --force \
             || pnpm publish --no-git-checks
+          cd ../..
+
+          cd typegraph/deno/sdk
+          ghjk x build-jsr-pub
+
+          # Requires auth
+          # jsr requires explicit type hints everytime (almost no inferences)
```
Yohe-Am, line 199: Yeah, not a fan of that requirement even though it makes total sense. Ghjk will need a lot of work to get there.

.vscode/settings.json
```diff
@@ -8,9 +8,13 @@
     "typegate",
     "dev",
     "examples/templates/deno",
+    "examples/typegraphs",
+    "typegraph/deno/dev",
+    "typegraph/deno/sdk",
     "ghjk.ts",
     ".ghjk"
   ],
+  "deno.importMap": "typegate/import_map.json",
```
Yohe-Am, line 17: 🙌🏿 hope this solves the IDE issues I've been suffering from

Yohe-Am (approved):

PR #770, opened 2024-06-28
feat(cli): list subcommand

j03-dev:
-

-

#### Migration notes

...
## Add `list` subcommand to List typegraph(s) from typegate

- [x] Change Something on CONTRIBUTING.md `fix contributing`


CONTRIBUTING.md
```diff
@@ -133,7 +133,7 @@ ghjk x dev-compose all # or only the envs required (e.g. base prisma s3)
 ghjk x test-e2e # all tests
```
Yohe-Am, line 133: Care to add that missing `build-graph` step here?

meta-cli/src/cli/list.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+impl List {
+    async fn get_more_info(&self) -> Result<Vec<TypegraphObject>> {
+        let client = reqwest::Client::new();
+
+        let mut headers = HeaderMap::new();
+        headers.insert(
+            AUTHORIZATION,
+            HeaderValue::from_str(&format!("Basic {}", "YWRtaW46cGFzc3dvcmQ="))?, // I don't know if it right, may be get from env
```
Yohe-Am, line 90: thought: you can get this info from the config. You just `base64(`${user}:${password}`). Or better yet, use the `post` method on the `Node` type found in the `common` crate. Grep for other usages in the cli for examples.

Yohe-Am (approved):

CONTRIBUTING.md
j03-dev: ok, done!

meta-cli/src/cli/list.rs
```diff
@@ -0,0 +1,175 @@
...
+            query {
+                typegraphs {
+                    name
+                    url
+                }
+            }
+        "#;
+
+        let response = client
+            .post("http://localhost:7891/typegate") // this too, may be get from env
```
destifo, line 103: I think you can also get this from the [config](https://github.com/metatypedev/metatype/blob/1d62c0320fb275ce9efabf772de2d604780105d7/examples/metatype.yaml#L3C5-L3C33),

destifo (commented):
great work all around 🙌

meta-cli/src/cli/list.rs
j03-dev:
```rust
impl Default for NodeConfig {
    fn default() -> Self {
        NodeConfig {
            url: "http://localhost:7890".parse().unwrap(),
            prefix: None,
            username: None,
            password: None,
            env: HashMap::default(),
            secrets: HashMap::default(),
        }
    }
}
```

the url is not same, the graphql is on 7891,

meta-cli/src/cli/list.rs
j03-dev: I already change it, but I wait @Natoandro to check why path in Typegraph is None


```rust
    async fn get_more_info(&self, dir: PathBuf) -> Result<Vec<TypegraphObject>> {
        let query = r#"
            query {
                typegraphs {
                    name
                    url
                }
            }
        "#;
        let mut node_config = NodeConfig::default();
        node_config.url = Url::parse("http://localhost:7891")?;
        let node = node_config.build(dir).await?;
        let response = node
            .post("/typegate")
            .unwrap()
            .gql(query.into(), None)
            .await?;

        response
            .data("typegraphs")
            .map_err(|err| anyhow::anyhow!(err))
    }

```

meta-cli/src/cli/list.rs
afmika: I would suggest you try change your local config to point at `localhost:7891` for testing purposes instead, the more generic we are the better it is. Or maybe use something similar to [this](https://github.com/metatypedev/metatype/blob/main/typegate/tests/e2e/templates/templates_test.ts#L50) for quick testing.

dev/tasks-install.ts
```diff
@@ -9,13 +9,18 @@ const tasks: Record<string, DenoTaskDefArgs> = {
...
-      if (/(Ubuntu|Debian)/.test(uname)) {
+      const osRelease = await $`cat /etc/os-release`.text();
+      if (/(Ubuntu|Debian)/.test(osRelease)) {
         console.log(
           `sudo apt update && ` +
             `sudo apt install -y --no-install-recommends ` +
             `gcc-multilib pkg-config libssl-dev libclang-dev perl make`,
         );
+      } else if (/Fedora|Red Hat|CentOS/.test(osRelease)) {
+        console.log(
```
zifeo, line 20: 👍

meta-cli/src/cli/list.rs
```diff
@@ -0,0 +1,175 @@
...
+    #[tracing::instrument]
+    async fn run(&self, args: ConfigArgs) -> Result<()> {
+        let dir = args.dir()?;
+        let config_path = args.config.clone();
+
+        let config = Arc::new(Config::load_or_find(config_path, &dir)?);
+
+        let console = ConsoleActor::new(Arc::clone(&config)).start();
+
+        let action_generator = SerializeActionGenerator::new(
```
zifeo, line 37: We will likely have other similar "simple command", do you see how we could simplify that setup into some utils to keep the run DRY?

meta-cli/src/cli/list.rs
```diff
@@ -0,0 +1,175 @@
...
+        format!(
+            "{:<50} {:<20} {:<30}",
+            self.name,
+            self.path.as_deref().unwrap_or("None"),
+            self.url.as_deref().unwrap_or("None")
+        )
+    }
+
+    fn print_table_header() {
+        println!("{:<50} {:<20} {:<30}", "NAME", "PATH", "URL");
```
zifeo, line 173: Some extensions (as this is your first exploratory PR, later tasks shall be fully described from the start) that may be worth implementing :
- add a column to specific which under which target the typography is deployed
- the `local` target could refer to the dev instance
- the target could be `-` for a non deployed tg

zifeo (commented):
Nice first contrib! Do not forget to sign the CLA and rebase your code :)

meta-cli/src/cli/list.rs
```diff
@@ -0,0 +1,175 @@
...
+            let mut table = Table::new(tg.name().unwrap());
+            table.set_path(tg.get_path().ok());
+            tables.push(table);
+        }
+
+        if let Ok(more_info) = self.get_more_info(dir).await {
+            for tg in more_info {
+                let mut table = Table::new(tg.name);
+                table.set_url(Some(tg.url));
+                tables.push(table);
```
Natoandro, line 126: /Merge entries with the same typegraph name?

meta-cli/src/cli/list.rs
```diff
@@ -0,0 +1,175 @@
...
+        for table in tables {
+            println!("{}", table.to_table_row());
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone)]
+struct Table {
```
Natoandro, line 140: I think you meant `TableRow` or `TableEntry` here.

typegraph/node/sdk/src/tg_manage.ts
```diff
@@ -26,6 +26,9 @@ export class Manager {
       case "deploy":
         await this.#deploy();
         break;
+      case "list":
+        await this.#list();
```
Natoandro, line 30: This won't be called as you use the `SerializeActionGenerator`...

Natoandro (changes requested):
Great !

meta-cli/src/cli/list.rs
Natoandro: I think we need to add the same `--target` flag as in the serialize subcommand.

typegraph/node/sdk/src/tg_manage.ts
j03-dev: ah, ok thanks

meta-cli/src/cli/list.rs
j03-dev: No, I just create table to display them

PR #771, opened 2024-06-29
test: Test typegate upgrade

Natoandro:
Test the typegate upgrade from the latest published version.


typegate/tests/e2e/upgrade/upgrade_test.ts
```diff
@@ -0,0 +1,264 @@
...
+  }
+}
+
+// download the fat version of the cli on the latest stable release
+async function downloadAndExtractAsset(version: string) {
+  const name = getAssetName(version);
+  const extractTargetDir = tempDir.join(name);
+  const metaBin = extractTargetDir.join("meta");
+  if (await checkMetaBin(metaBin)) {
+    return metaBin.toString();
```
Yohe-Am, line 78: The path here might resolve to a global `meta` installation. Try executing the file at the path instead of manipulating path vars? You can use $.raw`` if you're having issues with dax.

Yohe-Am (changes requested):

destifo (approved):
Great 👍

typegate/tests/e2e/upgrade/upgrade_test.ts
Natoandro: I have Deno permission issues with that...

Yohe-Am: Hmm. I wonder if `allow-run` takes globs.

PR #787, opened 2024-07-09
fix: Skip deno stack trace from error message

Natoandro:
Skip the deno stack trace from the error message when tg_manage fails.

#### Migration notes

_N/A_


zifeo (approved):

destifo (approved):

destifo:
we should really add the strict deno fmt though 😅

PR #792, opened 2024-07-16
chore: bump v0.4.5

Yohe-Am:
- Bumps metatype version to 0.4.5
- Bumps ghjk to latest commit
- Fixes `setup` whiz task to avoid issues on macos
- Fixes release pipeline to publish JSR

MET-614 MET-606 MET-605 MET-613

#### Migration notes

_No changes required._


.github/workflows/release.yml
```diff
@@ -71,36 +71,41 @@ jobs:
...
           # FIXME: deno doesn't support musl today https://github.com/denoland/deno/issues/3711
           # - os: ubuntu-latest
           #   target: x86_64-unknown-linux-musl
           #   suffix: ""
+          #   cross: false
           - os: ubuntu-latest
             target: aarch64-unknown-linux-gnu
             suffix: ""
+            cross: true
     steps:
```
afmika, line 92: Just curious, what's the main advantage of `cross`?

afmika (dismissed):

.github/workflows/release.yml
Yohe-Am: It allows us to build arm64 binaries on x86_64.

dev/consts.ts
```diff
@@ -1,16 +1,16 @@
 // Copyright Metatype OÜ, licensed under the Elastic License 2.0.
 // SPDX-License-Identifier: Elastic-2.0
 
-export const GHJK_VERSION = "0.2.0";
+export const METATYPE_VERSION = "0.4.5";
+export const PUBLISHED_VERSION = "0.4.4";
```
Natoandro, line 5: When we bump to a stable version, should not the published version be bumped to the same version? Because, it usually means that the version will be published soon...

Natoandro (dismissed):
:+1:

dev/consts.ts
Yohe-Am: Wouldn't that break the tests in this PR?

Natoandro (approved):

PR #799, opened 2024-07-24
feat:  better arg split logic for MCLI_LOADER

j03-dev:
-

-

#### Migration notes

...

### DIFF
--- a/meta-cli/src/deploy/actors/task.rs
+++ b/meta-cli/src/deploy/actors/task.rs
@@ -397,3 +397,135 @@ impl<A: TaskAction + 'static> Handler<Stop> for TaskActor<A> {
         }
     }
 }
+
+mod shell_words {
+    use core::fmt;
+
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    pub struct ParseError;
+
+    impl fmt::Display for ParseError {
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            f.write_str("missing closing quote")
+        }
+    }
+
+    #[cfg(feature = "std")]
+    impl std::error::Error for ParseError {}
+
+    enum State {
+        Delimiter,
+        Backslash,
+        Unquoted,
+        UnquotedBackslash,
+        SingleQuoted,
+        DoubleQuoted,
+        DoubleQuotedBackslash,
+        Comment,
+    }
+
+    pub fn split(s: &str) -> Result<Vec<String>, ParseError> {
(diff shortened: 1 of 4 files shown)


meta-cli/src/deploy/actors/task.rs
```diff
@@ -397,3 +397,132 @@ impl<A: TaskAction + 'static> Handler<Stop> for TaskActor<A> {
         }
     }
 }
+
+mod shell_words {
+    use core::fmt;
+
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    pub struct ParseError;
```
afmika, line 405: You should move this into a separate file and add the license on top, like [this one](https://github.com/metatypedev/metatype/blob/main/typegraph/deno/sdk/src/deps/_import.ts#L1)

meta-cli/src/deploy/actors/task/command.rs
```diff
@@ -78,8 +79,8 @@ impl CommandContext {
                 .to_string();
             let command = if !argv_str_fmtd.eq(&argv_str) {
                 // custom
-                let argv = argv_str_fmtd.split(' ').collect::<Vec<_>>();
-                let mut command = Command::new(argv[0]);
+                let argv = shell_words::split(&argv_str_fmtd).unwrap();
+                let mut command = Command::new(argv[0].clone());
```
afmika, line 83: `split` may fail as the env variable is provided by the user, so maybe you should add `context` to it before unwrap or  propagate the error if the function returns a result.

afmika (commented):
Looking good so far, just make sure you find all places where a naive `.split(' ')` is used on meta cli, the ones I provided was for reference.

meta-cli/src/deploy/actors/task.rs
afmika: yes, it's using the MIT license so it should be roughly the same as the example I provided.

meta-cli/src/utils/shell_words.rs
```diff
@@ -0,0 +1,153 @@
+// no-auto-license-header
+
+/*
+The MIT License (MIT)
+
+Copyright (c) 2020 Alexandre Piel
+Copyright 2018-2022 the Deno authors.
```
afmika, line 7: Wrong license? copy/paste the integrality of the author's license file content in here, then it should be ready to go 👍

afmika (changes requested):

afmika (approved):

PR #819, opened 2024-08-21
feat:  grpc runtime

j03-dev:
#### Migration notes

...

- [X] End-user documentation is updated to reflect the change


typegraph/core/src/runtimes/grpc/mod.rs
```diff
@@ -46,3 +39,30 @@ impl MaterializerConverter for GrpcMaterializer {
         })
     }
 }
+
+fn get_gprc_data(runtime_id: RuntimeId) -> Rc<GrpcRuntimeData> {
+    match Store::get_runtime(runtime_id).unwrap() {
+        Runtime::Grpc(data) => data,
+        _ => unreachable!(),
+    }
+}
```
michael-0acf4, line 48: Minimizing unwrap would be best imo, and instead always return Results (especially if assertions can break like in this case), wasm panics are pretty much unreadable at the time being.

typegraph/core/src/runtimes/grpc/type_generation.rs
```diff
@@ -0,0 +1,56 @@
...
+    let input_fields = get_message_field_descriptor(&file_descriptor, &input_message)?;
+    let output_fields = get_message_field_descriptor(&file_descriptor, &out_message)?;
+
+    Ok(Type {
+        input: traduct(input_fields)?,
+        output: traduct(output_fields)?,
+    })
+}
+
+fn traduct(fields: Fields) -> Result<TypeId> {
```
Natoandro, line 41: What does "traduct" mean?

typegraph/core/src/runtimes/grpc/type_generation.rs
```diff
@@ -0,0 +1,56 @@
...
+    let mut r#type = t::struct_();
+    for field in fields {
+        let the_type = match field.proto().type_name() {
+            "string" => t::string().build()?,
+            "int32" => t::integer().build()?,
+            "int64" => t::integer().build()?,
+            "bool" => t::boolean().build()?,
+            "float" => t::float().build()?,
+            tt => panic!("{tt} is not type"),
```
Natoandro, line 50: `Result::Err`?

Natoandro (approved):
Great work 💪 

PS: If you find any way to lighten the addition of new runtimes, feel free to write an RFC ;-)

Natoandro:
Can you update the description?

Natoandro (changes requested):
Tests required!

typegraph/core/src/runtimes/grpc/type_generation.rs
j03-dev: yeah i will change it ! still have strange error,

typegraph/python/typegraph/runtimes/grpc.py
```diff
@@ -0,0 +1,33 @@
...
+)
+from typegraph.gen.types import Err
+from typegraph.wit import runtimes, store
+
+
+class GrpcRuntime(Runtime):
+    def __init__(self, proto_file: str, endpoint: str):
+        file = open(proto_file, "r")
+        proto_file_content = file.read()
+        file.close()
```
michael-0acf4, line 18: I suggest you to rely on [fs.rs](https://github.com/metatypedev/metatype/blob/main/src/typegraph/core/src/utils/fs.rs#L123), this way you don't have to fight over workdir normalization accross the SDKs (something like [this](https://github.com/metatypedev/metatype/blob/7a221f62cef4dce2f9f2f961f0a5a855134ac22c/src/typegraph/core/src/utils/postprocess/deno_rt.rs#L23).

libs/common/src/grpc/proto_parser.rs
```diff
@@ -0,0 +1,167 @@
...
+            .trim_matches(';');
+        file_descriptor.set_package(package.to_string());
+        Ok(())
+    }
+
+    fn parse_message_start(line: &str) -> Result<DescriptorProto> {
+        let name = line
+            .split_whitespace()
+            .nth(1)
+            .ok_or_else(|| anyhow!("Invalid message line"))?
```
michael-0acf4, line 90: The message line is invalid, but which one? How can the user debug that?

libs/common/src/grpc/proto_parser.rs
```diff
@@ -0,0 +1,167 @@
...
+        message.set_name(name.to_string());
+        Ok(message)
+    }
+
+    fn parse_message_field(message: &mut DescriptorProto, line: &str) -> Result<()> {
+        let parts: Vec<&str> = line.split_whitespace().collect();
+        if parts.len() < 4 {
+            return Err(anyhow!("Invalid field line"));
+        }
+
```
michael-0acf4, line 102: Same here..

libs/common/src/grpc/proto_parser.rs
```diff
@@ -0,0 +1,167 @@
...
+    fn parse_enum_value(enum_type: &mut EnumDescriptorProto, line: &str) -> Result<()> {
+        let parts: Vec<&str> = line.split('=').collect();
+        if parts.len() != 2 {
+            return Err(anyhow!("Invalid enum value line"));
+        }
+
+        let name = parts[0].trim().to_string();
+        let number: i32 = parts[1].trim().trim_matches(';').parse()?;
+
```
michael-0acf4, line 153: What if there is a `/* multiline comment */` in between the number and `;`?

src/typegraph/deno/src/runtimes/grpc.ts
```diff
@@ -0,0 +1,24 @@
...
+import { runtimes } from "../wit.ts";
+import { Runtime } from "./mod.ts";
+
+export class GrpcRuntime extends Runtime {
+  constructor(protoFile: string, endpoint: string) {
+    const protoFileContent = new TextDecoder().decode(
+      Deno.readFileSync(protoFile),
+    );
+    const id = runtimes.registerGrpcRuntime({
+      protoFileContent,
```
michael-0acf4, line 14: I would suggest you to give the file path in the sdk and use fs context to read its content, unormalized `cwd` can give issues or introduce sneaky bugs.

michael-0acf4: Or better, treat proto files as artifacts just like how we do with DenoRuntime or PythonRuntime.

michael-0acf4 (changes requested):
I sort of get why you went the parser way but handling things like import resolution for example is not an easy task. Adapting [protobuf_parser ](https://docs.rs/protobuf-parse/latest/protobuf_parse/) to fit our needs might solve this. wdyt?

common.proto
```proto
syntax = "proto3";

package example;

message User {
    string name = 1;
    int32 id = 2;
}
```
main.proto
```
syntax = "proto3";

package example;

import "common.proto"; // <= here

service UserService {
    rpc listAll(User) returns (User);
}

```

src/typegraph/deno/src/runtimes/grpc.ts
j03-dev: But can we use fs context inside engine ? I'm not sure.

j03-dev: Definitively, we need to discuss .XD

docs/metatype.dev/docs/concepts/features-overview/index.mdx
```diff
@@ -59,9 +59,17 @@ We're taking any and all kinds of feature requests, suggestions and recommendati
       },
       {
         title: "gRPC API",
-        status: "future",
+        status: "complete",
         link: "/docs/reference/typegate",
-        body: <>Support to expose the typegraph through a gRPC API.</>,
+        body: (
+          <>
+             Expose your typegraph through a gRPC API, enabling high-performance,
```
Natoandro, line 66: > Expose your typegraph through a gRPC API
I think that is not accurate here, _"expose your gRPC API (...)"_ would be the one.

docs/metatype.dev/docs/reference/runtimes/grpc/index.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+### `proto_file`
+
+The `proto_file` is the path to the `.proto` file that defines the gRPC service. This file contains the service definitions, including message types and method signatures, that the Typegraph will use to communicate with the gRPC server.
+
+### `endpoint`
+
+The `endpoint` is the address of the gRPC server that the Typegraph will communicate with. It uses the format `tcp://<host>:<port>`, and is specified to point to the correct server and port where the gRPC service is running.
+
+### `call_grpc_method`
```
Natoandro, line 26: And this one is not a variable either.

Natoandro, line 26: Can we rename this to `call`? It is method default on the `GrpcRuntime` so it is already clear that it is going to call a gRPC method.

docs/metatype.dev/docs/reference/runtimes/grpc/index.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+This method is used to call a specific method on the gRPC service. It accepts the full path to the gRPC method, usually in the form `/package_name.service_name/method_name`. The **package_name** refers to the package defined in the `.proto` file, and it must be included when calling the method. In the example below, `greet` will call the `SayHello` method of the `Greeter` service within the `helloworld` package, as defined in the `helloworld.proto` file.
```
Natoandro, line 28: This method does not call a method on the gRPC service. Instead, it creates a typegraph function for a gRPC method call 😛

tests/runtimes/grpc/__snapshots__/grpc_test.ts.snap
```diff
@@ -0,0 +1,295 @@
...
+        "name": "deno",
+        "data": {
+          "worker": "default",
+          "permissions": {}
+        }
+      },
+      {
+        "name": "grpc",
+        "data": {
+          "proto_file_content": "syntax = \\\\"proto3\\\\";\\\\n\\\\npackage helloworld;\\\\n\\\\nservice Greeter {\\\\n  rpc SayHello (HelloRequest) returns (HelloReply) {}\\\\n}\\\\n\\\\nmessage HelloRequest {\\\\n  string name = 1;\\\\n}\\\\n\\\\nmessage HelloReply {\\\\n  string message = 1;\\\\n}\\\\n",
```
Natoandro, line 114: The proto file content could be large so, I suggest to upload it as an artifact instead of inlining. It is not required on this PR though, so just create a ticket for that.

tests/runtimes/grpc/geography.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+BASE_DIR = Path(__file__).parent
+
+
+@typegraph()
+def geography(g: Graph):
+    endpoint = "tcp://localhost:4770"
+
+    geography = BASE_DIR.joinpath("proto/geography.proto")
```
Natoandro, line 17: Paths in a typegraph definition are always relative to the parent directory. You could resolve the absolute path in the typegraph core.

tests/runtimes/grpc/helloworld.py
```diff
@@ -0,0 +1,23 @@
+# Copyright Metatype OÜ, licensed under the Elastic License 2.0.
+# SPDX-License-Identifier: Elastic-2.0
+
+from pathlib import Path
+
+from typegraph import Graph, Policy, typegraph
+from typegraph.runtimes.grpc import GrpcRuntime
+
+
+BASE_DIR = Path(__file__).parent
```
Natoandro, line 10: Idem

tests/runtimes/grpc/helloworld.ts
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: Elastic-2.0
+
+import { Policy, typegraph } from "@typegraph/sdk/index.ts";
+import { GrpcRuntime } from "@typegraph/sdk/runtimes/grpc.ts";
+
+const __dirname = new URL(".", import.meta.url).pathname;
+
+export const tg = await typegraph("helloworld", (g) => {
+  const endpoint = "tcp://localhost:4770";
+  const helloworld = `${__dirname}proto/helloworld.proto`;
```
Natoandro, line 11: Idem

tools/compose/proto/helloworld.proto
Natoandro, line 1: You could use a symbolic link here to prevent discrepancies when the test proto files are changes.

Natoandro (changes requested):
Great job 👍

michael-0acf4 (dismissed):
Can be merged as a first iteration👍
lgtm

docs/metatype.dev/docs/reference/runtimes/grpc/index.mdx
j03-dev: Ok, inderstand

michael-0acf4 (approved):

examples/typegraphs/grpc.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@typegraph(
+    # skip:start
+    cors=Cors(allow_origin=["https://metatype.dev", "http://localhost:3000"]),
+    # skip:end
+)
+def grpc(g: Graph):
+    endpoint = "tcp://localhost:4770"
+    proto_file = "typegraphs/proto/helloworld.proto"
```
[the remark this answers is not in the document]
j03-dev: but proto file is not read by typegraph, core

Natoandro (approved):
Great :+1:

Waiting for the tests to pass...

PR #820, opened 2024-08-25
chore: use stable deno std from jsr

Natoandro:
- Bump deno std version, import from jsr

-

#### Migration notes

_N/A_


zifeo (dismissed):

Yohe-Am (commented):
Oh wow, I'd actually done some of this work in my flatten deps PR though this PR seems more extensive. Will try to apply the diffs in the other PR (#821) but I think we should avoid merging this to avoid...pain lol

PR #822, opened 2024-08-26
refactor(docs, CLI): change to use meta-dev

destifo:
## ensure documentation is pushing for meta dev instead of meta typegate
- [ ] add a warning that envs are not set.

-

#### Migration notes

...


typegate/src/config/types.ts
```diff
@@ -32,21 +32,45 @@ export type GlobalConfig = z.infer<typeof globalConfigSchema>;
...
-  }),
+  tg_secret: z
+    .string()
+    .optional()
+    .transform((s: string | undefined, ctx) => {
+      if (s === undefined) {
+        console.error(
+          "\x1b[31m%s\x1b[0m",
+          "Error: Env var TG_SECRET is not configured. Exiting from the application.",
+        );
```
michael-0acf4, line 43: Maybe use `red(..)` from `deps.ts`?

typegate/src/config/types.ts
```diff
@@ -32,21 +32,45 @@ export type GlobalConfig = z.infer<typeof globalConfigSchema>;
...
+  tg_admin_password: z
+    .string()
+    .optional()
+    .transform((s: string | undefined, _ctx) => {
+      if (s === undefined) {
+        console.error(
+          "\x1b[31m%s\x1b[0m",
+          "Error: Env var TG_ADMIN_PASSWORD is not configured. Exiting from the application.",
+        );
+        Deno.exit(1);
```
michael-0acf4, line 69: I would advise against effects on validation, how about `ctx.addIssue({ ... })`?

destifo: that works but the output on the console is kind of verbose(especially when there are multiple issues), I can change that to only show the issue on top of the stack and elevate the verbosity... that's why I chose `console.error()`, wdyt though? change how we display those Zod validation errors?

Yohe-Am (dismissed):

michael-0acf4 (approved):

PR #840, opened 2024-09-06
feat: extend typegraph_core

luckasRanarison:
I'm still not convinced about the actual use cases but this is a draft of how `typegraph_core` could be used as a normal rust library by reusing wit generated types which are just normal Rust types and without exposing ABI related functions.

It exposes some type related utils like the `t` module which was previously private and is quite similar to the one in python and typescript. Though there are currently no wrapper around runtimes.

This is how it currently looks like: https://gist.github.com/luckasRanarison/26ceb1ed3152cadddf802a1f449c369d

#### Migration notes

...

### DIFF
--- a/src/typegraph/core/Cargo.toml
+++ b/src/typegraph/core/Cargo.toml
@@ -3,8 +3,9 @@ name = "typegraph_core"
 version = "0.4.11-rc.0"
 edition = "2021"
 
-[lib]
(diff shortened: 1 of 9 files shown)


Natoandro (commented):
Great 💪

luckasRanarison:
Closing this in  favor of truly separated crates. I'll submit an upcoming PR

PR #843, opened 2024-09-09
fix(cli): default installation directory

luckasRanarison:
Changed the default installation location to `~/.metatype/bin` and added prompt and hints for adding the directory to `PATH`.

#### Migration notes

...


installer.sh
```diff
@@ -102,24 +102,72 @@ Set the OUT_DIR environment variable to change the installation directory:
...
+  read -p "Do you want to append the new PATH to your configuration ($SHELL_CONFIG)? (y/n): " answer
+
+  answer=$(echo "$answer" | tr "[:upper:]" "[:lower:]")
+
+  case $SHELL_TYPE in
+    bash|zsh|ksh)
+      APPEND_CMD="export PATH=\"$OUT_DIR:\$PATH\""
+      ;;
+    fish)
+      APPEND_CMD="set -gx PATH $OUT_DIR \$PATH"
```
zifeo, line 150: you can use fish_add_path

installer.sh
```diff
@@ -102,24 +102,72 @@ Set the OUT_DIR environment variable to change the installation directory:
...
+$ $APPEND_CMD
+EOF
+  fi
+else
+  OUT_DIR=$(realpath $OUT_DIR)
+  if [[ ":$PATH:" != *":$OUT_DIR:"* ]]; then
+    cat <<EOF
 
 The installation directory is not in your PATH, consider adding it:
 $ export PATH="\$PATH:$OUT_DIR"
```
zifeo, line 170: should this be changed/removed?

installer.sh
```diff
@@ -102,24 +102,72 @@ Set the OUT_DIR environment variable to change the installation directory:
...
     mv "$TMP_DIR/$EXE" "$OUT_DIR"
   else
-    printf "Sudo is required to run \"sudo mv %s %s\":\n" "$TMP_DIR/$EXE" "$OUT_DIR"
-    sudo mv "$TMP_DIR/$EXE" "$OUT_DIR"
+    echo "$OUT_DIR is not writable."
+    exit 1
   fi
 fi
 
 rm -r "$TMP_DIR"
```
zifeo, line 118: should we remove the old path?

luckasRanarison: I think the executable should stay in a writable space, and we should avoid the default install to `/local/usr/bin`.

zifeo: no definitely, I meant deleting today's default path, so next time people run the installer, they get everything correct?

luckasRanarison: Oh, it should override the existing binary without any issue.

installer.sh
```diff
@@ -16,7 +16,7 @@ LATEST_VERSION="${LATEST_VERSION##*v}"
 
 PLATFORM="${PLATFORM:-}"
 TMP_DIR=$(mktemp -d)
-OUT_DIR="${OUT_DIR:-/usr/local/bin}"
+OUT_DIR="${OUT_DIR:-$HOME/.metatype/bin}"
```
Yohe-Am, line 19: Might I recommend the XDG directories instead? The binaries would go into `~/.local/bin` instead. Any other state would go to `~/.local/` or  `~/.config`. @zifeo wdyt?

zifeo: XDG convention may indeed be even better - but might be hard to check if on path already (not standard on Mac)?

Yohe-Am: XDG bin does seem to be ill defined on Mac. Let's keep with `.metatype` then.

Natoandro (approved):

michael-0acf4 (approved):

PR #844, opened 2024-09-09
feat(subs,gate): substantial integration

michael-0acf4:
Supersedes #818
* Backend constructs
* Workflow logic
  * Exceptions/Interrupts + configurable `SUBSTANTIAL_RELAUNCH_MS`
  * Runs (start, stop, healthcheck queries)
* Workflows in typescript for now
* Events (receive, handle)
* Std interrupts (`ctx.sleep(..)`, `ctx.ensure(..)`)

Also removed Substantial syntactic sugar from the Deno/PythonRuntime SDKs as it might introduce confusion on the inner workings of the Typegate.

#### Migration notes

None


src/substantial/src/backends/fs.rs
```diff
@@ -0,0 +1,177 @@
...
+
+impl FsBackend {
+    pub fn new(root: &Path) -> Self {
+        for d in &["runs", "schedules", "leases"] {
+            fs::create_dir_all(root.join(d))
+                .with_context(|| "Failed to create directory")
+                .unwrap();
+        }
+        Self {
+            root: root.to_owned(),
```
michael-0acf4, line 25: You might notice the missing `leases` on the backends, ressource management is done on the typegate and deno workers does most of the heavy lifting for channeling the events in a native non-blocking way (scheduling is done out of the box). If there are some use cases that I am missing please do tell me.

michael-0acf4: Reintroducing it wouldn't be that difficult I imagine but it would imply spawning our own task queue, and run an infinite async loop that triggers `Start` for leased workflows but I am not really sure were such loop would be placed at or when such loop would be initialized (in a custom worker maybe?).

src/typegate/src/runtimes/substantial.ts
```diff
@@ -0,0 +1,412 @@
...
+  }
+
+  #delegate(mat: TypeMaterializer): Resolver {
+    const matName = mat.name;
+    const data = mat?.data ?? {};
+    const workflowName = data.name as string;
+
+    switch (matName) {
+      case "start":
+        return this.#startResolver(workflowName);
```
michael-0acf4, line 159: The current implementation immediately starts the workflow. The poc implementation would put this in a schedule timeline registry-ish, and then an infinite loop would poll for the scheduled workflows in time order but it's all done automatically with workers.
Caveat being that `Start` information is lost if the typegate dies (we lose `resume` feature), so I am currently considering of adding it unless you have suggestions.

src/typegate/src/runtimes/substantial.ts
```diff
@@ -0,0 +1,412 @@
...
+            case "WAIT_ENSURE_VALUE":
+            case "WAIT_HANDLE_EVENT":
+            case "WAIT_RECEIVE_EVENT": {
+              await this.#workflowHandleInterrupts(workflowName, runId, ret);
+              break;
+            }
+            default: {
+              this.#workflowHandleGracefullCompletion(
+                startedAt,
+                workflowName,
```
michael-0acf4, line 322: Note: This is analoguous to the try block on the poc and happens when a run is complete (with errors, with a result or with interrupts).

src/typegate/src/runtimes/substantial/worker.ts
```diff
@@ -0,0 +1,68 @@
...
+
+self.onmessage = async function (event) {
+  const { type, data } = event.data as WorkerData;
+  switch (type) {
+    case "START": {
+      const { modulePath, functionName, run, kwargs } = data;
+      const module = await import(modulePath);
+      // TODO: for python use the same strategy but instead call from native
+      const workflowFn = module[functionName];
+
```
michael-0acf4, line 17: Idea being that for python, it would still in this worker file but instead we run a function from native that loads the artifact and calls the workflow `def`, the `Context` will be also written in python (or rust).

src/substantial/src/backends/fs.rs
michael-0acf4: Marking this as resolved, will extend the `WorkerManager` to handle leases.

src/typegate/src/runtimes/substantial.ts
michael-0acf4: Also marking this as a resolved question, workflow runs will be resumed if the typegate dies and restarted.

src/typegate/engine/00_runtime.js
```diff
@@ -51,4 +51,8 @@ globalThis.Meta = {
     destroy: getOp("op_wit_wire_destroy"),
     handle: getOp("op_wit_wire_handle"),
   },
+  substantial: {
+    createOrGetRun: getOp("op_create_or_get_run"),
```
Yohe-Am, line 55: nitpick: prefix the ops with `op_substantial` for namespacing?

src/typegate/engine/bindings.ts
```diff
@@ -454,3 +456,43 @@ export function archive(a0: ArchiveInp): ArchiveResult {
...
+  | {
+      Ok: string;
+    }
+  | {
+      Err: {
+        message: string;
+      };
+    };
+
+export function createOrGetRun(inp: CreateOrGetInput): CreateOrGetResult {
```
Yohe-Am, line 482: thought: these binding functions aren't exactly necessary. Feel free to use the methods on the Meta object directly. Should work as long as the types in runtime.d.ts are good.

src/typegate/engine/src/lib.rs
```diff
@@ -135,6 +136,11 @@ pub async fn launch_typegate_deno(
     main_mod: deno_core::ModuleSpecifier,
     import_map_url: Option<String>,
 ) -> Result<()> {
+    std::env::var("REDIS_URL")
```
Yohe-Am, line 139: This is probably a merge conflict re-add. We no longer use that variable.

src/typegate/engine/src/runtimes/substantial.rs
```diff
@@ -0,0 +1,92 @@
...
+
+#[derive(Clone, Default)]
+pub struct Ctx {
+    pub backends: Arc<DashMap<String, Box<dyn Backend>>>,
+}
+
+fn get_backend(kind: &SubstantialBackend) -> Box<dyn Backend> {
+    match kind {
+        // TODO: generalize
+        SubstantialBackend::Fs => Box::new(FsBackend::new(Path::new("tmp"))),
```
Yohe-Am, line 27: Highlighting that todo here. Also, our TMP_DIR is in `.metatype` now.

src/typegraph/core/src/runtimes/substantial.rs
```diff
@@ -23,6 +23,8 @@ pub enum SubstantialMaterializer {
     Start { workflow: WorkflowMatData },
     Stop { workflow: WorkflowMatData },
     Send { workflow: WorkflowMatData },
+    Ressources { workflow: WorkflowMatData },
```
Yohe-Am, line 26: Typo.

Yohe-Am (approved):
Looking good. But as the complexity increases, we'll need to invest in even more massive test infra. A sound model can still have faulty implementations.

src/typegate/engine/00_runtime.js
Yohe-Am: Resolved in #845.

src/typegate/engine/bindings.ts
Yohe-Am: Resolved in #845.

tests/runtimes/substantial/substantial.py
```diff
@@ -0,0 +1,38 @@
...
+
+    g.expose(
+        pub,
+        start=save_and_sleep.start(t.struct({"a": t.integer(), "b": t.integer()})),
+        ressources=save_and_sleep.query_ressources(),
+        results=save_and_sleep.query_results(
+            t.either([t.integer(), t.string()]).rename("ResultOrError")
+        ),
+        start_email=sub_email.start(t.struct({"to": t.email()})),
+        send_confirmation=sub_email.send(t.boolean()),
```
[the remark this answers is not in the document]
michael-0acf4: yes on the graphql side, but I will update this to something like `sub_email.start(event="confirmation", t.struct({"to": t.email()}))`) on #845 so it locks on a single event.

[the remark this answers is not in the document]
michael-0acf4: But at the same time, injections sort of make sense here..

src/typegraph/core/src/runtimes/substantial.rs
michael-0acf4: Ressources => Ressource?

michael-0acf4: Ah. nvm.

tests/runtimes/substantial/substantial.py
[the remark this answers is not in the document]
Yohe-Am: Injections are a great fit.

Natoandro (approved):

PR #854, opened 2024-09-24
feat(metagen/client): Add file upload support

Natoandro:
Solve [MET-629](https://linear.app/metatypedev/issue/MET-629/client-add-support-for-file-uploads)
- [x] `client_ts`
- [ ] `client_py`
- [ ] `client_rs`

#### Migration notes

_N/A_


tests/metagen/typegraphs/sample/ts_upload/client.ts
```diff
@@ -0,0 +1,878 @@
...
+    });
+  }
+    
+  upload(args: RootUploadFnInput | PlaceholderArgs<RootUploadFnInput>) {
+    const inner = _selectionToNodeSet(
+      { "upload": args },
+      [["upload", nodeMetas.RootUploadFn]],
+      "$q",
+    )[0];
+    return new MutationNode(inner, [[".file"]]) as MutationNode<RootUploadFnOutput>;
```
Yohe-Am, line 868: I'd recommend putting this metadata, the object paths, on the NodeMeta objects. You might have functions that take file arguments and if you collect them here, they'll be duplicated.

tests/metagen/typegraphs/sample/ts_upload/client.ts
```diff
@@ -0,0 +1,878 @@
...
+  return out;
+}
+
+/* Query node types section */
+
+type SelectNode<_Out = unknown> = {
+  nodeName: string;
+  instanceName: string;
+  args?: NodeArgs;
+  subNodes?: SelectNode[];
```
Yohe-Am, line 133: I rec collecting the found files from the `NodeMeta`s here. This would happen in _selectionToNodeSet`. You'd definitely prefer this approach in the Rust implementation which is way more strict.

In fact, I recommend doing the rust impl first since it's restrictions will provide more guidance.

tests/metagen/typegraphs/sample/ts_upload/client.ts
```diff
@@ -0,0 +1,878 @@
...
+      out = `${out} { ${
+        subNodes.map((node) => this.#convertQueryNodeGql(node, [])).join(" ")
+      } }`;
+    }
+    return out;
+  }
+
+  static build(
+    typeToGqlTypeMap: Record<string, string>,
+    query: Record<string, [SelectNode, ObjectPath[]]>,
```
Yohe-Am, line 461: You can then, while building the graphql query collect any files found in the `SelectNodes` same way we do args.

PR #856, opened 2024-09-24
feat: python hostcall

j03-dev:
may be work, need test

-

-

#### Migration notes

...

### DIFF
--- a/src/metagen/src/fdk_python/static/types.py.jinja
+++ b/src/metagen/src/fdk_python/static/types.py.jinja
@@ -2,6 +2,8 @@ from types import NoneType
 from typing import Callable, List, Union, get_origin, ForwardRef, Any
 from dataclasses import dataclass, asdict, fields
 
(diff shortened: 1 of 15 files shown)


src/metagen/src/fdk_python/static/types.py.jinja
```diff
@@ -96,7 +98,8 @@ def __repr(value: Any):
 def typed_{{ func.name }}(user_fn: Callable[[{{ func.input_name }}], {{ func.output_name }}]):
     def exported_wrapper(raw_inp):
         inp: {{ func.input_name }} = Struct.new({{ func.input_name }}, raw_inp)
-        out: {{ func.output_name }} = user_fn(inp)
+        ctx: {{ func.input_name}} = Ctx()
```
Natoandro, line 101: I think you should skip the type hint here, because it should be just `Ctx`

j03-dev: Okay

PR #860, opened 2024-09-30
feat: python hostcall

j03-dev:
Dead lock on python worker

-

-

#### Migration notes

...


tests/internal/py/logic.py
```diff
@@ -0,0 +1,19 @@
+# Copyright Metatype OÜ, licensed under the Elastic License 2.0.
+# SPDX-License-Identifier: Elastic-2.0
+
+
+def remote_sum(inp: dict, gql) -> float:
+    data = gql(
```
Yohe-Am, line 6: Try to emulate the existing custom function impls by passing in a ctx object instead of the `gql` function directly

Yohe-Am (dismissed):

Yohe-Am (approved):

PR #866, opened 2024-10-05
refactor!: Move as_id out of `TypeNode`

Natoandro:
Solve [MET-684](https://linear.app/metatypedev/issue/MET-684/store-the-id-field-in-tstruct-instead-of-in-the-target-type-as-id) and [MET-471](https://linear.app/metatypedev/issue/MET-471/parent-injection-use-property-name-instead-of-function-name)
- **common/typegraph**
  - [x] Store the id field in `ObjectTypeData` instead of in the target type (`as_id`)
  - [x] Add `id()` method on `t.integer` and `t.string`
- **typegraph/core**
  - [x] Store `as_id`, `injection` and `policy` in `TypeRef::attribute`
  - [x] Add support for direct and link target in `TypeRef`
  - [x] Only allow name registration for `TypeDef`
- Semantics
  - [x] Use property name instead of type name in from_parent injection source

#### Migration notes
**BREAKING CHANGE**
`from_parent` injections source shall be changed to the key in the parent `t.struct` instead of the type name.

#### Checklist


zifeo:
Very nice!

Yohe-Am (dismissed):
Great reduction of debt.

Yohe-Am:
I rec we release 0.5 after this is done.

PR #874, opened 2024-10-15
fix(ci): disable sccache when secrets not avail

Yohe-Am:
- Makes sccache optional so PRs from dependabot and forks can still run the test suite.
- Increases sccache allotment to 50g.


Yohe-Am:
Hmm, I wonder why the test suite is not triggering on this PR.

zifeo (approved):

PR #876, opened 2024-10-15
feat(gate): empty object as custom scalar

michael-0acf4:
* Allow empty object on the output without any change
* Just like `Int`, `String`, and such, rightfully refer the constant `{}` as a scalar
* **Any** empty object will now be refered as  `EmptyObject` scalar

#### Migration notes

None


Natoandro (approved):
Great!
There are a lot of formatting-related changes :thinking:

michael-0acf4:
> Great! There are a lot of formatting-related changes 🤔

yes, sorry for that. I believe it is because these files were never touched for after we dropped the deno formatter.

PR #892, opened 2024-10-28
feat: python workflows

j03-dev:
-

-

#### Migration notes

...


src/typegate/src/runtimes/substantial/worker.ts
```diff
@@ -1,60 +1,111 @@
...
+                    exception: wfException instanceof Error
+                      ? wfException
+                      : undefined,
+                    run: runCtx!.getRun(),
+                    schedule,
+                  } satisfies WorkflowResult,
+                ),
+              ),
+            );
+          });
```
michael-0acf4, line 70: I think you should use the same `then`/`catch` blocks for both.

src/typegate/src/runtimes/substantial/worker.ts
```diff
@@ -1,60 +1,111 @@
...
+      } else if (kind == "PYTHON") {
+        try {
+          const result = Meta.python.executePython({
+            python_module_path: modulePath,
+            python_function_name: functionName,
+            executing_context: runCtx,
```
michael-0acf4, line 76: Curious on how you will tackle the context methods. Will you try to unify it for both languages or.. ?

src/typegate/engine/Cargo.toml
```diff
@@ -68,6 +69,9 @@ bytes.workspace = true
 protobuf.workspace = true
 protobuf-json-mapping.workspace = true
 
+# python
+pyo3 = { workspace = true, features = ["extension-module"] }
```
michael-0acf4, line 73: > extension-module
nice!

michael-0acf4: By the way, which python version are you experimenting on this?
Is there a way to lock the Python version directly here?

michael-0acf4:
Suggestion:
Unifying the [context](https://github.com/metatypedev/metatype/blob/main/src/typegate/src/runtimes/substantial/deno_context.ts) for both languages into pure Rust should now be doable since we can use `Meta` inside workers (solves for the deno case), the python one will just use the rust implementation directly. The challenge would be on the hostcall.

src/typegate/engine/src/python.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+#[derive(Serialize)]
+#[serde(crate = "serde")]
+pub struct PythonExecutionOutput {
+    execution_result: Value,
+    execution_status: String,
+}
+
+fn convert_json_to_python_object(py: Python, json_value: &Value) -> PyResult<PyObject> {
```
zifeo, line 23: isn't pyo3 already providing such function?

src/substantial/src/converters.rs
```diff
@@ -125,6 +239,97 @@ impl Run {
...
+                            current_retry_count: None,
+                        });
+                    } else if let SavedValue::Retry {
+                        counter,
+                        wait_until,
+                    } = value
+                    {
+                        let now = Utc::now();
+                        if wait_until > &now {
+                            bail!(Interupt::Saveretry);
```
michael-0acf4, line 280: I suggest you to use enums directly. As we are within rust already, we can make clear distinctions between execution errors, interrupts and the returned value.

src/substantial/src/converters.rs
```diff
@@ -78,13 +79,126 @@ pub struct Operation {
...
+    Linear,
+}
+
+pub struct Retry {
+    pub strategy: Option<String>,
+    pub min_backoff_ms: i32,
+    pub max_backoff_ms: i32,
+    pub max_retries: i32,
+}
+
```
michael-0acf4, line 125: move into another file maybe? `converters.rs` should only contain conversion code imo.

src/substantial/src/converters.rs
```diff
@@ -125,6 +239,97 @@ impl Run {
     pub fn reset(&mut self) {
         self.operations = vec![];
     }
+
+    pub fn next_id(&mut self) -> u32 {
+        self.id += 1;
+        self.id
+    }
```
michael-0acf4, line 246: I suggest creating a context struct that owns a run and move that into another file, just like the way `deno_context.ts` is designed.

src/substantial/src/converters.rs
```diff
@@ -78,13 +79,126 @@ pub struct Operation {
 /// Each operation is produced from the workflow execution
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct Run {
+    pub id: u32,
```
Natoandro, line 82: Can you specify in a comment the difference between `id` and `run_id`?

PR #893, opened 2024-10-28
feat(metagen): add file upload support for Rust client

Natoandro:
Solve [MET-629](https://linear.app/metatypedev/issue/MET-629/client-add-support-for-file-uploads) (Part 1)
- Add file upload support to Rust metagen client

#### Migration notes

_N/A_


src/metagen/src/client_rs/static/client.rs
```diff
@@ -245,6 +247,191 @@ impl std::fmt::Display for SelectionError {
 }
 impl std::error::Error for SelectionError {}
 
+//
+// --- --- Input files --- --- //
```
Yohe-Am, line 251: nitpick: rec moving this section down below for better reading clarity.

src/metagen/src/shared/files.rs
```diff
@@ -0,0 +1,62 @@
...
+use common::typegraph::{
+    visitor::{Edge, PathSegment},
+    visitor2::{self, NearestFn, VisitNext},
+    Typegraph,
+};
+
+#[derive(Debug)]
+pub struct ValuePath(pub Vec<Cow<'static, str>>);
+
+fn serialize_path_segment(seg: &PathSegment) -> Result<Cow<'static, str>> {
```
Yohe-Am, line 13: Why did you decide against just outputting the ValuePathSegment literals instead of the parsed approach?

src/metagen/src/client_rs/static/client.rs
```diff
@@ -1250,175 +1439,423 @@ pub mod graphql {
...
-                        write!(dest, "(")?;
-                        for (key, val) in args {
-                            let name = format!("in{}", variable_types.len());
-                            write!(dest, "{key}: ${name}, ")?;
-                            variables_object.insert(name.clone(), val.value);
-                            variable_types.insert(name.into(), val.type_name);
-                        }
-                        write!(dest, ")")?;
+    #[derive(Debug)]
+    struct FileExtractor {
```
Yohe-Am, line 1443: nitpick: Maybe move this construct out of the `graphql` mod? It doesn't seem transport specific.

Yohe-Am (dismissed):

src/metagen/src/shared/files.rs
```diff
@@ -0,0 +1,62 @@
...
+        let mut path = Vec::with_capacity(tg_path.len());
+        for seg in tg_path {
+            path.push(serialize_path_segment(seg)?);
+        }
+        Ok(ValuePath(path))
+    }
+}
+
+pub fn get_path_to_files(tg: &Typegraph, root: u32) -> Result<HashMap<u32, Vec<ValuePath>>> {
+    visitor2::traverse_types(
```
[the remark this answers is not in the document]
Natoandro: This one is much simpler to use... I think we can remove the other one, it got somehow too complex...

michael-0acf4 (dismissed):
Nice!

.pre-commit-config.yaml
```diff
@@ -184,9 +184,12 @@ repos:
         exclude: tests/metagen/typegraphs/sample/rs/client.rs
       - id: clippy
         args:
+          - "--workspace"
           - "--locked"
           - "--all-features"
           - "--all-targets"
+          - "--exclude"
+          - "client_rs_static"
```
Yohe-Am, line 192: Note, we wan't to avoid triggering clippy in the user's workspace after the code is generated. Might be better to disable rules on the module directly if you need to instead of completely excluding it.

src/metagen/src/shared/files.rs
Yohe-Am: What I meant to say is that you get to avoid the parsing code if you just output literals directly. The `ValuePath` enum in the rust case.

Yohe-Am (approved):

src/metagen/src/shared/files.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use crate::interlude::*;
+use common::typegraph::{
+    visitor::{Edge, PathSegment},
+    visitor2::{self, NearestFn, VisitNext},
+    Typegraph,
+};
+
+#[derive(Debug)]
+pub struct TypePath(pub Vec<Cow<'static, str>>);
```
luckasRanarison, line 14: Could you make this non Rust specific and use types instead like before in order to share it between implementations.

luckasRanarison (changes requested):

luckasRanarison (approved):

PR #895, opened 2024-10-29
fix: meta dev does not exit properly upon `SIGTERM`

michael-0acf4:
#### Migration notes

None

### DIFF
--- a/src/meta-cli/src/deploy/actors/typegate.rs
+++ b/src/meta-cli/src/deploy/actors/typegate.rs
@@ -12,7 +12,6 @@ use rand::RngCore;
 use std::process::Stdio;
 use tokio::io::{AsyncBufReadExt, BufReader};
 use tokio::process::{ChildStderr, ChildStdout, Command};
-use tokio::sync::Mutex;
 
 pub mod message {
     use super::*;
@@ -30,7 +29,6 @@ pub struct TypegateActor {
     console: Addr<ConsoleActor>,
     #[allow(unused)]
     temp_dir: Option<tempfile::TempDir>,
-    child_process: Option<Arc<Mutex<tokio::process::Child>>>,
 }
 
 #[derive(Debug)]
@@ -94,7 +92,6 @@ impl TypegateInit {
             TypegateActor {
                 temp_dir: temp_dir_handle,
                 console,
-                child_process: None,
             }
         });
 
@@ -168,9 +165,7 @@ impl Handler<message::StartProcess> for TypegateActor {
         });
 
         match spawn_res {
-            Ok(Ok((child, stdout, stderr))) => {
-                self.child_process = Some(Mutex::new(child.into_inner()).into());
-
+            Ok(Ok((mut child, stdout, stderr))) => {
                 {
                     let addr = ctx.address();
                     let console = console.clone();
@@ -189,19 +184,14 @@ impl Handler<message::StartProcess> for TypegateActor {
                     ctx.spawn(fut.in_current_span().into_actor(self));
                 }
 
-                if let Some(child_arc) = &self.child_process {
-                    let addr = ctx.address();
-                    let child_arc = Arc::clone(child_arc);
-                    let fut = async move {
-                        let mut child = child_arc.lock().await;
-                        child
-                            .wait()
-                            .await
-                            .expect("failed to wait for typegate process");
-                        addr.do_send(message::Stop);
-                    };
-                    ctx.spawn(fut.in_current_span().into_actor(self));
-                }
+                let addr = ctx.address();
+                let fut = async move {
+                    Box::into_pin(child.wait())
+                        .await
+                        .expect("failed to wait for typegate process");
+                    addr.do_send(message::Stop);
+                };
+                ctx.spawn(fut.in_current_span().into_actor(self));
             }
 
             Err(e) | Ok(Err(e)) => {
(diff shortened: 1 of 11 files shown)


michael-0acf4:
A few remarks so far,
- This is not an issue with Ctrl+C or `SIGINT`
- This happens when we send a `SIGTERM` or `SIGKILL` to the parent process `meta dev`, `meta typegate` is still pending: making next cli spawn error out saying that the port is already used. So I imagine the issue is apparent in programmatic spawn context. I suspect there is something blocking that cannot be gracefully stopped with these.

Natoandro (dismissed):
Do you think you can add some tests?

michael-0acf4:
> Do you think you can add some tests?

So, I digged a bit and as far as I know, there seems to be no way to retrieve all PID of the child processes that are spawned by child processes in Deno.

Spawn, kill with SIGTERM/SIGHUP, spawn again then checking the std output for `Addr in use` is an alternative but I wonder if it's really worth it as it may came from bad port allocation and not the child process being alive itself...

luckasRanarison:
> So, I digged a bit and as far as I know, there seems to be no way to retrieve all PIDs or get handles of the child processes that are spawned by child processes in Deno.
Can't you just check the processes using shell commands?

Yohe-Am (approved):
Looks good. So, there's this [`FIXME`](https://github.com/metatypedev/metatype/blob/bb42c00169a34c4282317d50800993eea781a69d/tests/e2e/cli/dev_test.ts#L297) case in the tests where the `meta dev` leaks the `typegate` process. It doesn't seem to show up on my dev machine anymore but it was a regular on the coder machines. It also regularly leaked on the CI machines and...

...yeah, I just checked by running the CI on this PR and it still leaks it seems. The following screenshot is from [this](https://github.com/metatypedev/metatype/actions/runs/11601023885/job/32302817163) CI run observed through a tmate session.
image
The two tests that use the `meta-full` build have already passed by the time I took this screenshot.
Hopefully, this PR improves the state in developer shells but the bug seems still at large.

michael-0acf4:
> Looks good. So, there's this [`FIXME`](https://github.com/metatypedev/metatype/blob/bb42c00169a34c4282317d50800993eea781a69d/tests/e2e/cli/dev_test.ts#L297) case in the tests where the `meta dev` leaks the `typegate` process. It doesn't seem to show up on my dev machine anymore but it was a regular on the coder machines. It also regularly leaked on the CI machines and...
> 
> ...yeah, I just checked by running the CI on this PR and it still leaks it seems. The following screenshot is from [this](https://github.com/metatypedev/metatype/actions/runs/11601023885/job/32302817163) CI run observed through a tmate session.
> image
> The two tests that use the `meta-full` build have already passed by the time I took this screenshot.  
> Hopefully, this PR improves the state in developer shells but the bug seems still at large.
> 
> Edit: replaced screenshot with one that shows currently running test.

That is surprising, the way I tested it was sending the signals directly through `htop`, and kill the parent `cargo run -p meta-cli -- dev` for SIGTERM. I also experimented quitting the terminal that spawned it for the SIGHUP case (+htop) so it's all manual.. not sure if there is something that I might have missed but ultimately i will try to write a test for it after all.

michael-0acf4:
> there's this FIXME
Ctrlc would be SIGKILL for that matter and that at least should work even without this patch.

Yohe-Am:
> Ctrlc would be SIGKILL for that matter and that at least should work even without this patch.

Ctrl C corresponds to SIGTERM actually. SIGKILL is the harsher cousin that immediately kills the process. You're not able to register handlers for it and do clean up. In that FIXME, I'm using SIGTERM to give it a chance to clean up but no bueno.

michael-0acf4:
> > Ctrlc would be SIGKILL for that matter and that at least should work even without this patch.
> 
> Ctrl C corresponds to SIGTERM actually. SIGKILL is the harsher cousin that immediately kills the process. You're not able to register handlers for it and do clean up. In that FIXME, I'm using SIGTERM to give it a chance to clean up but no bueno.

My bad, I meant SIGINT. But you are right, SIGTERM is more appropriate.

src/meta-cli/src/deploy/actors/typegate.rs
```diff
@@ -206,6 +216,14 @@ impl Handler<message::Stop> for TypegateActor {
     type Result = ();
 
     fn handle(&mut self, _msg: message::Stop, ctx: &mut Self::Context) {
+        if let Some(child_arc) = self.child_process.take() {
+            let fut = async move {
+                let mut child = child_arc.lock().await;
+                let _ = child.kill().await;
```
Natoandro, line 222: Not sure if I understand correctly, but looks like the mutex will be locked in the code above until the child process exits.
> ```
> let mut child = child_arc.lock().await;
> child
>   .wait()
>   .await
>   .expect("failed to wait for typegate process");
> addr.do_send(message::Stop);
> ```

So the lock would only be acquired here after the child process has stopped...

Natoandro (dismissed):

src/meta-cli/src/deploy/actors/typegate.rs
michael-0acf4: Good catch, in fact none of this are required now. The [Signal hook](https://docs.rs/signal-hook/latest/signal_hook/) overrides most of the logic.

luckasRanarison (approved):

tests/e2e/cli/dev_test.ts
```diff
@@ -263,114 +271,145 @@ Meta.test(
...
+        ["rate.ts", "rate"],
+        ["reduce.ts", "roadmap-reduce"],
+        ["rest.ts", "roadmap-rest"],
+        ["roadmap-policies.ts", "roadmap-policies"],
+        ["roadmap-random.ts", "roadmap-random"],
+        ["triggers.ts", "triggers"],
+      ]);
+    });
+
+    await t.should("kill meta-full process", async () => {
```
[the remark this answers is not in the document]
michael-0acf4: Not really sure but I suspect there is an issue on how the signals propagates in test context, I was not able to pinpoint exactly where though.
What I find really interesting is that I cannot reproduce it the behavior when directly using the binary, even if we say, quit the terminal that spawned a meta dev (which is the most likely thing to happen during development), the signals still propagate without issue.
But the good thing is that ctrlc works as expected in tests.

PR #898, opened 2024-10-31
fix: speculative fix for `typegate_prisma_test.ts`

Yohe-Am:
- Assigns special schemas for each test that relies on the `runtimes/prisma/prisma.py` typegraph.
- Bumps version tag to 0.5.0-rc4
- Fixes issues in release pipeline
- Disables all but the `test-full` job when the test pipeline is run with tmate enabled. 

The flakeout of `typegate_prisma_test.ts` has proved difficult to recreate. Looking at the code and going from recent similar cases, I suspect it happens due to the old code reusing the same pg schema for multiple tests. Assigning special schemas for each tests should hopefully help.

#### Migration notes

N/A

---

### DIFF
--- a/tests/runtimes/typegate/__snapshots__/typegate_prisma_test.ts.snap
+++ b/tests/runtimes/typegate/__snapshots__/typegate_prisma_test.ts.snap
@@ -2,13 +2,320 @@ export const snapshot = {};
 
 snapshot[`typegate: find available operations 1`] = `
 {
-  errors: [
-    {
-      extensions: {},
-      locations: [],
-      message: "'config' not found at 'FindPrismaModels.findPrismaModels.fields.type', available names are: optional, title, type, enum, default, format, policies",
-      path: [],
-    },
-  ],
+  data: {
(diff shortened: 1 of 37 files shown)


Natoandro (approved):
Different schema name is definitely the right solution...
Good job :muscle:

tests/runtimes/typegate/__snapshots__/typegate_prisma_test.ts.snap
```diff
@@ -2,13 +2,320 @@ export const snapshot = {};
...
 snapshot[`typegate: find available operations 1`] = `
 {
-  errors: [
-    {
-      extensions: {},
-      locations: [],
-      message: "'config' not found at 'FindPrismaModels.findPrismaModels.fields.type', available names are: optional, title, type, enum, default, format, policies",
-      path: [],
-    },
-  ],
```
Natoandro, line 12: This is why we should avoid large PRs :hand_over_mouth:

PR #904, opened 2024-11-04
fix(typegraph): send rpc message in chunks in the TS typegraph client 

Natoandro:
- Send the JSON-RPC message is chunks in the TypeScript typegraph client to prevent reaching the line size limit for stdout. Note: we could not reproduce the issue locally as it only occurs when using the published package for Node.js.
- Use JSON-RPC notification for logging and report from the typegraph clients.
- Other changes:
  - Use relative paths for static task sources in the CLI;
  - Fix TODO in `meta gen`: pass the working directory on the `working_dir` param of `SerializeActionGenerator::new`.

#### Migration notes
_N/A_

---

### DIFF
--- a/src/typegraph/deno/src/io.ts
+++ b/src/typegraph/deno/src/io.ts
@@ -133,7 +133,7 @@ const rpcCall = (() => {
       params,
     });
 
-    process.stdout.write(`jsonrpc: ${rpcMessage}\n`);
+    writeRpcMessage(rpcMessage);
     return responseReader.read(rpcId);
   };
 })();
--- a/src/typegraph/deno/src/io.ts
+++ b/src/typegraph/deno/src/io.ts
@@ -10,8 +10,8 @@ import process from "node:process";
 const JSONRPC_VERSION = "2.0";
 
 function writeRpcMessage(message: string) {
-  // split into 32-kibibyte chunks
-  const chunkSize = 32768;
+  // split into 32-KiB chunks
+  const chunkSize = 32758; // 32 KiB - 10 bytes for "jsonrpc^: " or "jsonrpc$: "
   for (let i = 0; i < message.length; i += chunkSize) {
     const chunk = message.slice(i, i + chunkSize);
     if (i + chunkSize < message.length) {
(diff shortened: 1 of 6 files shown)


src/typegraph/deno/src/io.ts
```diff
@@ -7,6 +7,38 @@ import process from "node:process";
  * see: module level documentation `meta-cli/src/deploy/actors/task.rs`
  */
 
+const JSONRPC_VERSION = "2.0";
+
+function writeRpcMessage(message: string) {
+  // split into 32-KiB chunks
+  const chunkSize = 32758; // 32 KiB - 10 bytes for "jsonrpc^: " or "jsonrpc$: "
+  for (let i = 0; i < message.length; i += chunkSize) {
+    const chunk = message.slice(i, i + chunkSize);
```
luckasRanarison, line 16: Shouldn't we encode the message first? Because we can definitely get variable length characters.

luckasRanarison: Actually nevermind, I forgot that you took a margin and that JavaScript correctly handles UTF-8.

luckasRanarison (approved):

src/typegraph/deno/src/io.ts
michael-0acf4, line 16: Since the message length returns the number of code units not the characters. With string slice there is still a risk of cutting in the middle of a character.
I suggest using `Array.from(message)` to get an actual iterator to the characters then perform slice on that. That would also mean additional fixes around the actual total length though.. but if it's about sending raw bytes it should be fine I guess.

```js
const text = "Hello👍";
console.log(text.length); // 7 not 6
console.log(text.slice(0, 6));
console.log(Array.from(text).slice(0, 6). join(""));
```

michael-0acf4 (approved):

src/typegraph/deno/src/io.ts
Natoandro: Good point... Each entry is parsed as a string in the rust size, so UTF validation might fail...
Unless the implementation follows this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#well-formed_json.stringify

PR #916, opened 2024-11-13
fix: Remove prisma count

Natoandro:
- Remove prisma count operation: it does not work, making it work like on the prisma client would complicate the prisma runtime (until we have output transformations...); Use aggregate instead.
- Increase the delay before exiting the process on the nodejs typegraph client to give the CLI time to process all the output.
- Remove `quaint` logs on the typegate (too verbose).

#### Migration notes

---

### DIFF
--- a/.ghjk/lock.json
+++ b/.ghjk/lock.json
@@ -1092,25 +1092,25 @@
               "ty": "denoFile@v1",
               "key": "dev-gate6",
               "desc": "Launch the typegate from a locally found meta bin.",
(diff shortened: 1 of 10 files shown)


michael-0acf4 (commented):
That is surprising, if I recall this was just a field like any other that we interfaced with the typegraph. Did it change at some point?

luckasRanarison (approved):

michael-0acf4 (approved):

PR #934, opened 2024-12-01
feat(metagen): typescript client file upload

luckasRanarison:
- Closes [MET-768](https://linear.app/metatypedev/issue/MET-768/add-file-upload-support-for-typescript-metagen-client).

#### Migration notes

---


src/metagen/src/client_ts/static/mod.ts
```diff
@@ -436,22 +543,43 @@ async function fetchGql(
...
+
+  const additionalHeaders: HeadersInit = {};
+
+  if (files && files.size > 0) {
+    const data = new FormData();
+    data.set("operations", body);
+    const map: Record<string, string[]> = {};
+    for (const [i, [path, file]] of [...(files?.entries() ?? [])].entries()) {
+      const key = `${i}`;
+      // TODO single file on multiple paths
```
luckasRanarison, line 561: By the way @Natoandro, what do you exactly mean by this?

Natoandro: If the same file is used multiple times, then it should not be emitted more than once in the request payload... Should be added to the array...

Yohe-Am (dismissed):

src/metagen/src/client_ts/static/mod.ts
luckasRanarison: Ah, I understand now why you used IDs for the Rust implementation.

michael-0acf4 (approved):

src/metagen/src/client_ts/static/mod.ts
luckasRanarison: I'll make a follow up PR for the python one and add more tests for the path extraction

Yohe-Am (approved):

PR #935, opened 2024-12-02
feat: injection on output types (outjection)

Natoandro:
- [x] Solve [MET-140](https://linear.app/metatypedev/issue/MET-140/gate-computestage-processing-from-context-should-evaluate-in-all-cases)
  - Enable injection on output
- [x] Solve [MET-47](https://linear.app/metatypedev/issue/MET-47/gate-from-parent-for-eitherunion)
  - test from parent injection for either/union types

#### Migration notes
- _N/A_

---


michael-0acf4 (dismissed):
Nice!

luckasRanarison (dismissed):

michael-0acf4 (dismissed):

tests/injection/outjection.py
```diff
@@ -0,0 +1,33 @@
...
+        Policy.public(),
+        randomUser=deno.identity(t.struct()).extend(
+            {
+                "id": t.uuid().from_random(),
+                "age": t.integer().set(19),
+                "email": t.email().from_context("user_email"),
+                "password": t.string().from_secret(
+                    "user_password"
+                ),  ## should we allow this?
+                "createdAt": t.datetime().inject("now"),
```
Yohe-Am, line 23: Does metagen handle this well btw? I.e. do we need to generate separate types without the injeted fields for the FDK/custom function implementers? I don't think it does, should I add it to the meagen papercuts tickeT?

tests/injection/outjection.py
```diff
@@ -0,0 +1,33 @@
...
+    random = RandomRuntime(seed=1)
+
+    g.expose(
+        Policy.public(),
+        randomUser=deno.identity(t.struct()).extend(
+            {
+                "id": t.uuid().from_random(),
+                "age": t.integer().set(19),
+                "email": t.email().from_context("user_email"),
+                "password": t.string().from_secret(
```
Yohe-Am, line 20: tangent: do we have support or will we want to add support for optional injetions? I.e. only inject if no value is found on output?

tests/injection/outjection.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@typegraph()
+def outjection(g: Graph):
+    deno = DenoRuntime()
+    random = RandomRuntime(seed=1)
+
+    g.expose(
+        Policy.public(),
+        randomUser=deno.identity(t.struct()).extend(
```
Yohe-Am, line 15: tangent: would it be sensible to move `extend` onto `t.Struct`?

Yohe-Am (approved):

tests/injection/outjection.py
Natoandro: `extend` was originally on `t.struct`, but it is currently missing on the TS SDK. On `t.func`, it applies on the output type (if it is a `t.struct`). Main reason was the heavy use of `deno.identity` in the tests, which btw won't play well with injections after this PR...

tests/injection/outjection.py
Natoandro: Interesting... We can explore that...

tests/injection/outjection.py
Natoandro: For injection on output types ("outjections"), it think it is not needed. But, it is indeed necessary for injections..

tests/injection/outjection.py
Yohe-Am: https://linear.app/metatypedev/issue/MET-778/move-extend-to-tstructand-documentation

tests/injection/outjection.py
Yohe-Am: But wouldn't the type system complain if you don't provide these outjected values in your custom function responses?

tests/injection/outjection.py
Yohe-Am: https://linear.app/metatypedev/issue/MET-780/optional-injections

tests/injection/outjection.py
Natoandro: Ah, you are right... It is needed for the FDK...

luckasRanarison (approved):

PR #937, opened 2024-12-10
feat(gate,sdk)!: new policy spec

michael-0acf4:
#### Migration notes

* Replaced true, false, and null to ALLOW, DENY and PASS.

Composition rules:
1. On traversal order:

* `ALLOW`: allow parent and all its children (ignore inner policies)
* `DENY`: deny parent and all its children  (ignore inner policies)
* `PASS`: pass through parent and evaluate each children (no-op, equivalent to no policies)

2. On a single type (a.with_policy(X).with_policy(Y)):

`ALLOW` and `DENY` compose the same as true and false with the AND gate, `PASS` does not participate.
* `ALLOW` & P = P
* `DENY` & P = `DENY` (e.g. DENY & ALLOW = DENY)
* `PASS` & P = P (does not participate)

## Release Notes

- **New Features**
	- Enhanced documentation for Metatype's mental model, including clearer policy definitions and a comparison table with classical models.
	- Introduction of a comprehensive tutorial on building a Metatype API, covering setup, CRUD operations, and security practices.

- **Bug Fixes**
	- Updated policy logic to return explicit 'ALLOW' or 'DENY' strings instead of boolean values across various components.

- **Documentation**
	- Improved clarity and detail in documentation for policies and core concepts.
	- Added new sections for policy composition rules and traversal order.

- **Refactor**
	- Streamlined policy management and evaluation logic across multiple files, enhancing clarity and maintainability.

- **Tests**
	- Added tests for new policy functionalities and updated existing tests to reflect changes in policy handling.


michael-0acf4:
Note:

* 98% of the old policies tests were not modified as they do not involve complex nesting.
* For the new changes [tests/policies/policies_composition_test.ts](https://github.com/metatypedev/metatype/blob/policy-spec-impl/tests/policies/policies_composition_test.ts) should give more context.

Yohe-Am (approved):

Yohe-Am:
Nice, this will need a docs update though.

michael-0acf4:
@coderabbitai review

examples/typegraphs/roadmap-policies.ts
```diff
@@ -50,7 +50,7 @@ typegraph(
 
     const admins = deno.policy(
       "admins",
-      "(_args, { context }) => !!context.username",
+      "(_args, { context }) => !!context.username ? 'ALLOW' : 'DENY'",
```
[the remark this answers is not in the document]
michael-0acf4: Not required

tests/policies/policies_composition.py
```diff
@@ -0,0 +1,87 @@
...
+from typegraph.runtimes import DenoRuntime
+
+
+@typegraph()
+def policies_composition(g: Graph):
+    deno = DenoRuntime()
+
+    def ctxread(pol_name: str):
+        return deno.policy(pol_name, code=f"(_, {{ context }}) => context.{pol_name}")
+
```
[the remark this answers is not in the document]
michael-0acf4: Not required, it will fail if that was undefined.

docs/metatype.dev/docs/reference/policies/index.mdx
```diff
@@ -28,3 +28,33 @@ Policies are hierarchical in the sense that the request starts with a denial, an
...
+If you have `foo.with_policy(A, B).with_policy(C)` for example, it will be merged into a single chain `[A, B, C]`.
+
+The evaluation is as follows:
+
+- `ALLOW` and `DENY` compose the same as `true` and `false` under the logical `AND` operator.
+- `PASS` does not participate.
+
+Or more concretely:
+
+- `ALLOW` & Other -> Other
```
Yohe-Am, line 51: This might be misleading, no? Doens't `ALLOW` completely short-circuit policy evaluation? I.e., if we get an allow, do we evaluate other members of the policy set on that node? I do like the analogy but maybe you should note this fact.

Yohe-Am: Also, we should not that ALLOW would be a very rare case. Most policies would not want to override any possible sibling policies. I'd avoid the alphanumeric ordering and list allow last to highlight this fact.

examples/typegraphs/execute.py
```diff
@@ -53,7 +53,7 @@ def roadmap(g: Graph):
     admins = deno.policy(
         "admins",
         """
-  (_args, { context }) => !!context.username
+  (_args, { context }) => !!context.username ? 'ALLOW' : 'DENY'
```
Yohe-Am, line 56: suggestion: we'd probably want `ALLOW` : `PASS` in the examples to encourage composable policies.

docs/metatype.dev/docs/reference/policies/index.mdx
michael-0acf4: >  Doens't ALLOW completely short-circuit policy evaluation

In traversal order yes. 

But since we allow multiple policies to be put on a single type I figured just `AND`-ing them would be more natural.

In fact, it follows the idea of...
```ts
struct(
  a: integer().with_policy(A),
  b: integer().with_policy(B)
)
```

The policies would be hoisted as `[A, B]`, and field access is just `A & B` if we assume we project all fields. Thus by extension...
```ts
struct(
  a: integer().with_policy(A),
  b: integer().with_policy(B, X)
)
```
We have `[A, B, X] := A & B & X`.

michael-0acf4: I have reformulated the explanation into something like "if one item on the chain denies then the entire thing denies".

Still the same idea but hopefully more simple to reason about.

E.g. `pseudo: t.string().with_policy(deny).with_policy(allow)` should deny `pseudo`

michael-0acf4: @Yohe-Am 

I have reformulated the explanation into something like "if one item on the chain denies then the entire thing denies".

Still the same idea but hopefully more simple to reason about.

E.g. `pseudo: t.string().with_policy(deny).with_policy(allow)` should deny..

Yohe-Am (approved):
nitpick: maybe too late but now I'm a bit confused around PASS and ALLOW naming lol. PASS is for PASSTHROUGH and not BYPASS right?

Edit: yeah, it is PASSTHROUGH.

PR #940, opened 2024-12-16
fix: prevent infinite recursion on visitor2

Natoandro:
- Solve [MET-781](https://linear.app/metatypedev/issue/MET-781/metagen-stack-overflow)
  - Skip the node when a cycle is detected

#### Migration notes

---


Natoandro:
@luckasRanarison I could not reproduce the issue with the local tests for cyclic types [here](https://github.com/metatypedev/metatype/blob/64ed2108e88e4d89cbc44c0085f6f4ab40e42b28/tests/metagen/metagen_test.ts#L105). It would be great if you could setup a small repro ;-p

luckasRanarison:
@Natoandro Yohe already fixed that issue in his PR so I think you can close this

PR #960, opened 2025-01-13
fix(cli): avoid expected error backtrace

luckasRanarison:
- Closes [MET-801](https://linear.app/metatypedev/issue/MET-801/avoid-backtraces-on-expected-errors).

#### Migration notes

---

## Summary by CodeRabbit

- **Chores**
	- Updated error reporting configuration to reduce verbosity during panic events.


michael-0acf4 (dismissed):

luckasRanarison:
@Yohe-Am Sorry, do we need some fancy output or is this enough?

Yohe-Am:
Hey, sorry but I found some more nuance on this ticket and had to change ghjk as well. So the panic section only seems to show up on explicit panics and not on `Err` values bubbled to main so my ticket screenshot is misleading. Otherwise, setting the flag discussed [here](https://github.com/eyre-rs/eyre/discussions/213) seems to be enough to avoid that tail hint.

Maybe there cli commands that aren't properly bubbling their errors? I wonder if we could enable more clippy lints for that or `unwrap`.

Natoandro (approved):

Yohe-Am (approved):

PR #1000, opened 2025-04-15
fix(sdk): node SDK cross-platform compatibility

luckasRanarison:
- Closes [MET-876](https://linear.app/metatypedev/issue/MET-876/nodejs-file-descriptor-api-not-working-on-macos).

- FFI had to be used to make file descriptors blocking on unix systems because they are asynchrnous by default, meaning a file descriptor read immediately fails when there's no data instead of waiting. Deno abstracts this in its implementation but we have to handle edge cases by ourselves with NodeJS.
- Fixes CI issues by upgrading ghjk to v0.2.2.

#### Migration notes

---

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved compatibility for reading standard input on macOS and Linux systems, ensuring smoother and more reliable input handling.
- **Chores**
	- Updated the version of the development environment tool used across workflows, documentation, and build configurations to v0.2.2.
- **Documentation**
	- Updated a Python code generation example in the documentation for clearer reference.


Natoandro (approved):

Yohe-Am (approved):

luckasRanarison:
No, it's just scoped to the current process standard input, Deno does it behind the scenes. But I think it's not really needed to bring back the flag because that's our desired behavior.

src/typegraph/specs/codegen/rpc/typescript/client.ts
```diff
@@ -10,6 +11,37 @@ const state = { id: 0 };
 const isDeno = !Deno.version.v8.includes("node");
 const encoder = new TextEncoder();
 const decoder = new TextDecoder();
+const { platform } = process;
+
+if (platform === "darwin" || platform === "linux") {
```
michael-0acf4, line 16: Why did you choose to also handle linux?

src/typegraph/specs/codegen/rpc/typescript/client.ts
```diff
@@ -10,6 +11,37 @@ const state = { id: 0 };
...
+
+  const { fcntl } = ffi.define({
+    fcntl: {
+      library: "libc",
+      retType: ffi.DataType.I32,
+      paramsType: [ffi.DataType.I32, ffi.DataType.I32, ffi.DataType.I32],
+    },
+  });
+
+  const flags = fcntl([process.stdin.fd, F_GETFL, 0]);
```
michael-0acf4, line 36: Worth to mention that stdin/out can be shared by the parent/child processes. So, wouldn't changing the blocking mode also affect the meta or something that runs meta? It might break some assumptions. The default is blocking after all.. even on linux or does the big guys like deno do the same?

luckasRanarison: Indeed if the stdio is inherited this also affects the parent's io behavior but it's completely safe because typegraph script processes are spawned with **piped** stdio. Also just a correction, it's async by default...

src/typegraph/specs/codegen/rpc/typescript/client.ts
luckasRanarison: Just to make the read process standardized and simpler

src/typegraph/specs/codegen/rpc/typescript/client.ts
michael-0acf4: > piped stdio
Got it.

> it's async by default...
Ubuntu 24.04
image

src/typegraph/specs/codegen/rpc/typescript/client.ts
michael-0acf4: I would suggest not doing that for linux's case + only do it if it's node (you set isDeno flag but did not use it?
The issue is only on mac after all. When something breaks again it can then be narrowed done quickly either wide or mac only.

michael-0acf4: I do get it for the case of testing though.. might not be important for the moment if all is working.

src/typegraph/specs/codegen/rpc/typescript/client.ts
michael-0acf4: wait. **non** block flag 0 is async?

michael-0acf4: >  const result = fcntl([process.stdin.fd, F_SETFL, flags & ~O_NONBLOCK]);

would no op on linux (if we ignore syscalls)

michael-0acf4: Same thing on mac, so wouldn't it just no op in all cases?

mac

michael-0acf4: @luckasRanarison ah, I think I get it.
1. It no ops on the blocking flag.
2. But it clears the over flags => so that made it work?

michael-0acf4: Sorry for the spam but I think I managed to narrow down the behavior out of curiosity.

For future references, it is platform + runtime specific only behavior (this is why I got no ops in all cases on Python3 even on MAC like the screenshots above).

1. flag is 0 on (Node, Linux) and (Deno, Linux), this is why it worked without any changes.
2. flag is 0 on (Deno, MAC) too, this is why it worked without any changes.
3. flag is 4 on (Node, MAC), and as such it does not no op but **do** reset it but...  

.. weirdly enough.. even though  flag is 4 on (Node, MAC), uncommenting the clear op ` const result = fcntl([process.stdin.fd, F_SETFL, flags /*& ~O_NONBLOCK*/])` which should be no op (so should EAGAIN) **makes it work** and removing this line makes it EAGAIN. That suggests either a bug on Node side or that that was retrieved is not the actual current default state.

PR #1008, opened 2025-05-02
fix(perf): Improve the performance of the typegraph expansion

Natoandro:
The previous single-pass recursive algorithm for the typegraph expansion was buggy and had terrible performance.
This PR attempts to simplify the expansion, making it easier to debug and have better performance. For more details, see the doc comments at the crate root (`src/typegraph/graph/src/lib.rs`).

Other changes:
- Change the github runners:
  - Downgrade to Ubuntu 22.04
  - Use the native Github runners for ARM

#### Migration notes

_N/A_

---

- [ ] ~End-user documentation is updated to reflect the change~

## Summary by CodeRabbit

- **New Features**
  - The tree view web tool now displays a dynamic footer status bar showing loading state and type count, and supports union/either node types in typegraph visualization. The default tree depth can be set via a URL parameter.

- **Bug Fixes**
  - Corrected naming and ordering of selection types and partial types in generated Rust, TypeScript, and Python client code for composite and cyclic structures.
  - Fixed a job naming typo in the test workflow.

- **Refactor**
  - Major internal overhaul of typegraph conversion logic, simplifying and restructuring the process for converting, linking, and registering types.
  - Streamlined and clarified logic for rendering TypeScript FDK templates and type registries.
  - Improved visibility and access to generated TypeScript client manifest types.

- **Chores**
  - Updated GitHub Actions workflows to use explicit Ubuntu 22.04 runners for consistency.
  - Increased verbosity of the meta-cli serialize command for tree view tooling.
  - Updated dependencies in configuration files.

- **Documentation**
  - Added comprehensive module-level documentation for the typegraph expansion process.

- **Tests**
  - Added and reorganized migration test files and improved structure of test-generated client code for clarity and accuracy.


src/metagen/src/client_ts/selections.rs
```diff
@@ -135,6 +135,10 @@ pub fn manifest_page(tg: &typegraph::Typegraph) -> ManifestPage<TsSelection> {
                 let ty_props = ty.properties();
                 let mut props = Vec::with_capacity(ty_props.len());
                 for (prop_name, prop) in ty_props {
+                    // TODO support for nested functions (MET-862)
+                    if matches!(prop.ty, Type::Function(_)) {
+                        continue;
```
Yohe-Am, line 140: We still get the void typed clients, right?

src/typegraph/graph/src/lib.rs
```diff
@@ -1,12 +1,34 @@
 // Copyright Metatype OÜ, licensed under the Mozilla Public License Version 2.0.
 // SPDX-License-Identifier: MPL-2.0
 
+//! The typegraph expansion creates a version of the typegraph where type
+//! references are inlined. This will remove the need to fetch from the array
+//! of types each time. Also, it will eventually duplicate some types
```
Yohe-Am, line 6: What is meant by duplication and inlining here? Using arcs instead of indices?

src/typegraph/graph/src/lib.rs
```diff
@@ -1,12 +1,34 @@
...
+//!
+//! The expansion works in 3 passes:
+//! - The first pass traverse the typegraph schema from the root using BFS.
+//!   Each step will convert the type if it has not been converted yet.
+//!   Also, the step will output the `LinkStep`, informing the system on how
+//!   should the created type be associated with others.
+//! - The second pass iterates through the created link steps. It sets the
+//!   references to child types. This pass is necessary since the child types
+//!   are not yet converted at each step. Recursive algorithm will not work
+//!   since the graph might have cycles.
```
Yohe-Am, line 18: Where are the performance wins coming from?

Yohe-Am (approved):

src/metagen/src/client_ts/selections.rs
Natoandro: Yes...

src/typegraph/graph/src/lib.rs
Natoandro: > inlining
Yes, arcs instead of indices, that way we won't need a reference to the typegraph each time we use a type.

> duplication
When the same type is used more than once, under different functions with different injections, it is duplicated because we have semantically different types.

Also, with the duplication, it is hard to keep the type references with indices...

src/typegraph/graph/src/lib.rs
Natoandro: I am not really sure. But the recursive algorithm requires a lot of checks at each node, due to the complexity, and trying to have the type complete after each node is traversed.

PR #1018, opened 2025-05-22
feat!: Optimize type count

Natoandro:
- Add optimization step for the typegraph: deduplicate types when possible.
- New UI for the tree-view-web; with lazy loading of subtrees [incomplete]

#### Migration notes

**BREAKING CHANGE**: Names are not taken into account when deduplicating, so it is possible that a type does not have the user-defined name. This is mostly relevant for metagen and the introspection.

---

## Summary by CodeRabbit

- **New Features**
  - Introduced a web-based UI for browsing and inspecting typegraphs, featuring type selection, hierarchical type navigation, and statistics display.
  - Added an API server with endpoints for listing typegraphs and retrieving type details.
  - Implemented a command-line flag to load typegraph data from JSON files.
  - Added a new optimization step for typegraphs, improving performance by deduplicating types and analyzing strongly connected components.

- **Improvements**
  - Enhanced type safety and reliability by replacing floating-point fields with non-NaN wrappers across multiple components.
  - Enabled hashing for several data types, improving support for hash-based operations.
  - Refined numeric validation logic to unwrap values before validation for consistency.

- **Bug Fixes**
  - Improved numeric validation logic for floating-point constraints.

- **Documentation**
  - Added comprehensive documentation and configuration files for the new frontend application.

- **Chores**
  - Updated dependencies and workspace configuration.
  - Added and updated ignore files to keep the repository clean.


src/typegraph/core/src/types/sdk/core.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright Metatype OÜ, licensed under the Mozilla Public License Version 2.0.
 // SPDX-License-Identifier: MPL-2.0
 
+use ordered_float::NotNan;
```
michael-0acf4, line 4: How can we get a nan?

michael-0acf4 (commented):
Nice work!

Worth to try probably is to have an utility that duplicates an existing typegraph N times, I believe we need to check any method with some or all of our example typegraphs when needs be.
Then we can keep control of the how much the deduplication is improving or degrading throughout iterations... and more importantly its reduction trend to check wether or not we hit a wall.

src/typegraph/core/src/types/sdk/core.rs
Natoandro: By passing NaN from JS, I guess...
This is just to enable hashing... NaN can have a lot of different underlying values...

michael-0acf4: I am guessing you got into that scenario this is why but, I am surprised we get a NaN out of a serde deserialized typegraph as it is not part of JSON spec.

Natoandro: I did not get NaN. It is required for `#[derive(Hash)]`...

PR #1027, opened 2025-06-05
fix: prisma query batching

Yohe-Am:
- Fixes prisma request batching support
- Bumps version to v0.5.1-rc.5

#### Migration notes

---

### DIFF
--- a/src/typegate/src/runtimes/prisma/prisma.ts
+++ b/src/typegate/src/runtimes/prisma/prisma.ts
@@ -158,39 +158,6 @@ export class PrismaRuntime extends Runtime {
     return ret;
   }
 
-  #executeResolver(
-    q: GenQuery,
-    path: string[],
-    renames: Record<string, string>,
-  ): Resolver {
-    return async ({ _: { variables, context, effect, parent } }) => {
-      path[0] = renames[path[0]] ?? path[0];
-
-      const startTime = performance.now();
-      const generatedQuery = q({ variables, context, effect, parent });
-      const res = await this.query(generatedQuery);
-      const endTime = performance.now();
-      this.logger.debug(
-        `queried prisma in ${(endTime - startTime).toFixed(2)}ms`,
-      );
-
-      if (path[0] == "queryRaw") {
-        const rawRes = res[0];
-        return rawRes.queryRaw.rows.map(
-          (row: any[]) =>
-            Object.fromEntries(
-              row.map(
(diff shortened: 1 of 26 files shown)


src/typegate/src/runtimes/prisma/prisma.ts
```diff
@@ -278,73 +253,125 @@ export class PrismaRuntime extends Runtime {
...
+    if (new Set(stageOrder).size != stageOrder.length) {
+      throw new Error("duplicate top stage ids");
+    }
 
-    fields.shift();
-    // TODO renames
+    const batchStage = new ComputeStage({
+      ...stage.props,
+      materializer: {
+        data: { operation: "batch" },
```
michael-0acf4, line 284: nitpick: Fake stage? If it's only for temporary compute then I suggest we prefix with fake, tmp, or similar.

michael-0acf4: I don't see why not use it although it does break the assumption that it comes from the query. Instead we can maybe create a new kind of stage specifically for this sort of purpose? We can all at least be clear that there are two kind of stages.

Yohe-Am: A new kind of shadow stage makes sense but I'd hold that off until we have more usecases. Abstraction count is harder to manage than code count. In this specific case though, I think I'll just use a bespoke function inline that will be cached or something. The introduction of this stage is breaking some of the tests.

src/typegate/src/runtimes/prisma/prisma.ts
```diff
@@ -278,73 +253,125 @@ export class PrismaRuntime extends Runtime {
...
+      };
+      stagesMat.push(
+        new ComputeStage({
+          ...field.props,
+          dependencies: [
+            ...field.props.dependencies,
+            ...(field.props.parent?.id() ?? "")
+              .split(".")
+              .filter((str) => str.length)
+              .map(str => parentToResolver.get(str)!)
```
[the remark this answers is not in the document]
Yohe-Am: I've removed the line

src/typegate/src/runtimes/prisma/prisma.ts
Yohe-Am: Keeping the batch stage turned out to be simpler than using a cached callback. It's no worse what what was being done before, where the first prisma function stage that was encountered was being used as a batch stage for other prisma functions.

luckasRanarison (approved):

PR #1028, opened 2025-06-05
fix(gate)!: auth refactor

luckasRanarison:
- Solves [MET-885](https://linear.app/metatypedev/issue/MET-885/auth-refactor).

#### Migration notes

---

## Summary by CodeRabbit

- **New Features**
  - OAuth2 authentication now supports multiple clients per provider with explicit client ID and redirect URI configuration.
  - Introduced a secure PKCE-based token flow requiring Redis for state management, including new token exchange and refresh endpoints.
  - Added Redis configuration support for both sync and non-sync modes.

- **Improvements**
  - Unified OAuth2 configuration across Python, TypeScript, and Rust SDKs with a single structured interface replacing provider-specific methods.
  - Enhanced validation and security for OAuth2 redirect URIs and client identification.
  - Updated JWT token duration defaults and made them configurable.
  - Modularized Redis URL parsing and connection logic.

- **Bug Fixes**
  - Improved error handling and validation in OAuth2 token and authorization flows.

- **Documentation**
  - Updated OAuth2 authentication docs to reflect the new PKCE token flow, token refresh, and Redis requirements.
  - Added Redis configuration and updated JWT duration defaults in environment variable documentation.

- **Tests**
  - Expanded authentication tests to cover PKCE, Redis state management, token exchange, and refresh flows.

- **Chores**
  - Updated example projects to use the new OAuth2 configuration format.
  - Removed deprecated OAuth2 helper methods and related code.


tests/auth/auth_test.ts
```diff
@@ -2,27 +2,53 @@
...
+  const data = encoder.encode(verifier);
+  const digest = await crypto.subtle.digest("SHA-256", data);
+  return base64url.encodeBase64Url(new Uint8Array(digest));
+}
+
+function generateCodeVerifier() {
+  const array = new Uint8Array(32);
+  crypto.getRandomValues(array);
+  return base64url.encodeBase64Url(new Uint8Array(array));
+}
```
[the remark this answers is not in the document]
luckasRanarison: encode and decode are deprecated

luckasRanarison:
I still need to update the docs and see how to integrate the scopes param with the custom profiler, but is what I'm doing ok so far?

src/typegate/src/services/auth/protocols/oauth2.ts
```diff
@@ -320,42 +373,42 @@ export class OAuth2Auth extends Protocol {
...
-    const payload: JWTClaims = {
+  async createJWT(request: Request, profile: Record<string, unknown>) {
+    if (this.authProfiler) {
+      profile = await this.authProfiler!.transform(profile, request);
+    }
+
+    const refreshAt = Math.floor(
+      new Date().valueOf() / 1000 + this.config.jwt_refresh_duration_sec,
+    );
+    const payload = {
```
Yohe-Am, line 395: suggestion: There are some [standard JWT claim keys](https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-token-claims#registered-claims) that we should be using.

src/typegate/src/services/auth/protocols/oauth2.ts
```diff
@@ -320,42 +373,42 @@ export class OAuth2Auth extends Protocol {
...
-      refreshAt: Math.floor(
-        new Date().valueOf() / 1000 +
-          (token.expiresIn ?? this.config.jwt_refresh_duration_sec),
-      ),
-      scope: token.scope,
+      refreshAt,
       profile,
     };
-    return await this.cryptoKeys.signJWT(
+    const access_token = await this.cryptoKeys.signJWT(
```
Yohe-Am, line 400: This is still using symmetric JWT signatures, right?

src/typegate/src/services/auth/protocols/oauth2.ts
```diff
@@ -176,24 +210,40 @@ export class OAuth2Auth extends Protocol {
...
+          request.headers,
+          this.typegraphName,
+          this.cryptoKeys,
+        );
+        const tokens = await client.code.getToken(url, {
+          state: state.provider.state,
+          codeVerifier: state.provider.codeVerifier,
+        });
+        const profile = await this.fetchProfile(tokens);
+        const token = await this.createJWT(request, profile);
```
Yohe-Am, line 226: The tokens should be minted at token swap, not during the authorization flow. The issued_at would be off otherwise. And I suppose we should set a time limit for how long the client can wait before the token swap.

src/typegate/src/services/auth/protocols/oauth2.ts
```diff
@@ -215,47 +265,68 @@ export class OAuth2Auth extends Protocol {
...
+      },
+      client: {
+        id: query.client_id,
+        state: query.state,
+        redirectUri: query.redirect_uri,
+        codeChallenge: query.code_challenge,
+        method: query.code_challenge_method,
+      },
+    };
+    const headers = await setEncryptedSessionCookie(
```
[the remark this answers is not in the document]
Yohe-Am: If we need to put some confidential information inside the code, like the IdP profile, encrypt them first.

examples/typegraphs/iam-provider.py
```diff
@@ -14,7 +14,12 @@
     # skip:end
 )
 def iam_provider(g: Graph):
-    g.auth(Auth.oauth2_github("openid profile email"))
+    g.auth(
+        Auth.oauth2_github(
```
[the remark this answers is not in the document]
luckasRanarison: > Not a fan of these provider specific methods
I also definitely think so, it's not really worth it

src/typegate/src/config/types.ts
```diff
@@ -104,6 +104,11 @@ export const typegateConfigBaseSchema = z.object({
   tmp_dir: z.string(),
   jwt_max_duration_sec: z.coerce.number().positive(),
   jwt_refresh_duration_sec: z.coerce.number().positive(),
+  redis_url: z
```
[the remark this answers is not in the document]
michael-0acf4: I think this should be fine and for now best is to simplify imo, redis can scale well horizontally. We already have an env for sub, but this on other hand is about storing state indefinitely.

michael-0acf4 (commented):
Neat!

I suggest we rely on the same register logic for sync mode, we want high availability without having to query redis when getting values each time but sync on set.

src/typegate/src/services/auth/protocols/oauth2.ts
luckasRanarison: For now...

examples/demo/demo.py
```diff
@@ -8,7 +8,18 @@
     rate=Rate(window_limit=2000, window_sec=60, query_limit=200),
 )
 def public_api(g: Graph):
-    g.auth(Auth.oauth2_github("openid profile email"))
+    g.auth(
+        Auth.oauth2(
+            provider="github",
+            scopes=["openid", "profile", "email"],
```
Yohe-Am, line 14: nitpick: name scopes can be bit unclear here since we now support scopes for the custom profiler as well.

src/typegate/src/services/auth/routes/token.ts
```diff
@@ -43,17 +60,26 @@ export async function token(params: RouteParams) {
...
     throw new Error("no redis connection");
   }
 
   try {
     if (body.grant_type === "authorization_code") {
-      const { token, state, code } = await getEncryptedCookie(
-        request.headers,
-        engine.name,
-        engine.tg.typegate.cryptoKeys,
+      const rawData = await redis.get(`code:${body.code}`);
```
Yohe-Am, line 69: nitpick: are these redis keys prefixed elsewhere? If we're sharing the URL, might be best to use more specifc prefixes. I use redis-commander a bunch and it's getting busy.

Yohe-Am (approved):
Looking great. Maybe we can one off try this with a standard client or a test suite? Just to shake out any auth issues once and for all (and make tickets if necessary).

examples/demo/demo.py
luckasRanarison: Good point, how should we handle it?

src/typegate/src/services/auth/routes/token.ts
luckasRanarison: I don't think they're used elsewhere

Yohe-Am (approved):

